PR #10, opened 2016-03-18
ApiSdk: Extended the response returned by the ApiSession to return Ap…

grastogi23:
…iResponse which provides a convenience function obj() that can be used to directly get object dictionary in all good cases. In case, response was HTTP 404 then it throws exception ObjNotFound. In all other error scenarios it throws Exception APIError


grastogi23:
@chaitanyaavi

PR #27, opened 2016-04-04
Default Objects, PKI profile ref issue, api-upload option updated

chaitanyaavi:
1) Default objects
2) PKI profile ref issue
3) api-upload option updated for added objects

### DIFF
--- a/python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
+++ b/python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
@@ -178,6 +178,8 @@ def get_defaults(f5_monitor, monitor_config, monitor_name):
     default objects and forms complete object
     :param f5_monitor: F5 monitor object
     :param monitor_config: List of F5 monitor configs
+    :param monitor_name: There is no attribute in config to determine type of
+    monitor it can be mapped from root monitors name
     :return:
     """
     parent_name = f5_monitor.get("defaults from", None)
@@ -201,6 +203,11 @@ def convert_monitor_entity(name, f5_monitor):
     :param f5_monitor: F5 monitor config object
     :return: Avi monitor config object
     """
+    reverse = f5_monitor.get("reverse", None)
+    if reverse:
+        parts = reverse.rsplit(" ")
+        f5_monitor[parts[0]] = parts[1]
+        f5_monitor["reverse"] = None
     supported_attributes = ["timeout", "interval", "time until up",
                             "description", "type", "defaults from"]
     skipped = [key for key in f5_monitor.keys()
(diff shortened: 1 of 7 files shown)


python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
```diff
@@ -183,7 +171,7 @@ def convert_pool_config(pool_config, tenant, monitor_config_list):
     return pool_list
 
 
-def get_defaults(f5_monitor, monitor_config):
+def get_defaults(f5_monitor, monitor_config, monitor_name):
```
grastogi23, line 174: why does it only take params like monitor_config and monitor_name. seems like function tries to find defaults for multiple types of objects. 

Also, let us split into smaller get_defaults as it is getting way too large.

python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
```diff
@@ -529,8 +520,8 @@ def convert_profile_config(profile_config, certs_location, option):
                 LOG.warning("Skipped PKI profile for profile %s "
                             "because of missing CA file" % name)
 
-            elif ca_file_name:
-                ca_file_name = ca_file_name.replace('\"', '')
+            elif ca_file_name and ca_file_name.replace('\"', '').strip():
```
grastogi23: do you need to check for both ca_file_name and ca_file_name.replace('\"', '').strip() you can simply do ca_file_name.replace('\"', '').strip():

python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
```diff
@@ -540,18 +531,18 @@ def convert_profile_config(profile_config, certs_location, option):
                     ca = ca_file.read()
                     pki_profile["ca_certs"] = [{'certificate': ca}]
                 except:
-                    LOG.error("Error to read file" +
+                    LOG.error("Error to read file " +
```
grastogi23: use format strings instead. like LOG.error("Error to read file %s%s%s", certs_location,os.path.sep,ca_file_name) this is more efficient and less error prone

python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
```diff
@@ -1,14 +1,11 @@
-import json
 import copy
 import numbers
 import logging
 import os
-
-from openpyxl import Workbook
+import csv
```
grastogi23, line 5: can you pl. check if we need to add this to the setup.py for the sdk. I believe csv is part of standard package but let us test it before we merge it in.

PR #29, opened 2016-04-06
External monitor and snat pool

chaitanyaavi:


python/avi/sdk/utils/f5_converter/f5_config_converter_v10.py
```diff
@@ -33,6 +33,16 @@ def get_port_by_protocol(protocol):
     return port
 
 
+def upload_file(file_path):
+    file_str = None
+    try:
+        file_obj = open(file_path, "r")
+        file_str = file_obj.read()
+    except:
+        LOG.error("Error to read file %s" % file_path)
```
grastogi23, line 42: If the upload file fails then the settings should be added in the skipped list.

PR #34, opened 2016-04-08
Sdk gr

grastogi23:


python/avi/sdk/samples/autoscale/aws_samplescaleout.py
```diff
@@ -0,0 +1,182 @@
...
+        boto.ec2.networkinterface.NetworkInterfaceCollection(interface)
+    print ami_id, interfaces
+    reservations = conn.run_instances(
+        ami_id, instance_type=instance_type, network_interfaces=interfaces)
+    if not reservations.instances:
+        return '', '', ''
+    instance = reservations.instances[0]
+    # Wait for the instance to enter the running state
+    # check for instance is running
+    while instance.update() != 'running':
```
adityats, line 105: maybe give up after some time, if the instance failed to come up (resource, configuration, etc.. issues)

python/avi/sdk/samples/autoscale/aws_samplescaleout.py
```diff
@@ -0,0 +1,182 @@
...
+    if not API:
+        API = ApiSession.get_session(
+                '127.0.0.1', 'admin', 'avi123', tenant='admin')
+    return API
+
+
+def create_aws_connection(aws_settings):
+    """
+    creates aws connection
+    :param aws_settings: dictionary of aws settings like credentials etc.
```
adityats, line 70: help string to better articulate the exact dictionary keys, if user is doing a help() on this method.

python/avi/sdk/samples/autoscale/aws_samplescaleout.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    aws_access_key_id = aws_settings['aws_access_key_id']
+    aws_secret_access_key = aws_settings['aws_secret_access_key']
+    ec2_region = aws_settings.get('ec2_region', 'us-west-2')
```
azizavi, line 74: Should the default region be 'us-west-2' or a failure? May be we can skip the boto call.

grastogi23, line 74: This sample code and I wanted to make it easy for anyone to try it out with minimal settings. If they don't have access to us-west-2 or use bad security groups then it would fail else we ask less questions to the user. 

This is code is merely sample with priority to become easy to understand. Full blown code with exception handling and all corner cases would make it hard for anyone to follow.

PR #37, opened 2016-04-13
AV-8511 #comment change to use logger from robot.api

hoangdinhini:


ypraveen:
This is going to be public API; using robot.api logging does not make sense here

grastogi23:
Hi Hoang,
As praveen mentioned. We cannot use robot in public SDK. Is there something
not working when you are using robot with standard logging? Just wondering
what prompted this change.

thanks
Gaurav

On Wed, Apr 13, 2016 at 7:32 AM, Praveen Yalagandula <
notifications@github.com> wrote:

> This is going to be public API; using robot.api logging does not make
> sense here
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> https://github.com/avinetworks/sdk/pull/37#issuecomment-209475999

## 

Gaurav Rastogi

grastogi AT avinetworks.com
Avi Networks

hoangdinhini:
Hi all,

Thanks for reviewing.
We have functional-metrics jobs on Jenkins showing logger.debug from avi_api to console output. This quite messes up the output. 
http://10.10.0.62:8080/job/eng-functional-metrics/2279/consoleFull
This change is to hide the the debug info.

If you have other ways, please let me know.

Thanks
Hoang

grastogi23:
when I am running the tests I don't see the prints... so there is something wrong in the way jenkins is setup. It must be enabling the robot debug / traces.

PR #47, opened 2016-04-28
Changes for indirect mapping column

chaitanyaavi:
Bug fixes


python/avi/f5_converter/f5_config_converter_v11.py
```diff
@@ -24,6 +24,16 @@ def upload_file(file_path):
     return file_str
 
 
+def update_skipped_attributes(skipped, indirect_list, ignore_dict, object):
+    indirect_mappings = [attr for attr in indirect_list if attr in skipped]
```
grastogi23, line 28: add comments

PR #55, opened 2016-05-16
F5 converter

chaitanyaavi:


python/avi/f5_converter/conversion_util.py
```diff
@@ -0,0 +1,167 @@
...
+        if "description" in src_cp:
+            del src_cp["description"]
+        del tmp_cp["name"]
+        if "description" in tmp_cp:
+          del tmp_cp["description"]
+        dup_lst = tmp_cp.pop("dup_of", [])
+        if cmp(src_cp, tmp_cp) == 0:
+            dup_lst.append(src_obj["name"])
+            tmp_obj["dup_of"] = dup_lst
+            return tmp_obj["name"]
```
grastogi23, line 60: it is only returning name then why keep tmp_obj['dup_of']

python/avi/f5_converter/f5_config_converter_v10.py
```diff
@@ -1,93 +1,15 @@
 import copy
-import numbers
 import logging
+import numbers
 import os
-import csv
+
+import conversion_util as util
```
grastogi23: uutil is very common name. we should name it to util. why don't we import explicit items from the conversion_util. or just use it as it is everywhere or call it conv_utils

python/avi/f5_converter/conversion_util.py
chaitanyaavi, line 60: tmp_obj['dup_of'] is used to map merged profiles if referred in VS returns name for logging and conversion status.later we remove it by remove_dup_key

PR #58, opened 2016-05-23
refactor the check to be more safe

squirrelala:
sorry for the double commit; just realized that the original way of checking num_instances first is safer (to protect against num_instances=None, since apparently None < 0 is True)

### DIFF
--- a/python/avi/sdk/utils/mesos/mesos_testutils.py
+++ b/python/avi/sdk/utils/mesos/mesos_testutils.py
@@ -186,13 +186,14 @@ class MesosTestUtils(object):
             app_obj = self.MARATHON_APP_TEMPLATES[app_type]
             app_obj = deepcopy(app_obj)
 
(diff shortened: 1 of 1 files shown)


python/avi/sdk/utils/mesos/mesos_testutils.py
```diff
@@ -186,13 +186,14 @@ def createApp(self, marathon_url, app_type, app_name, num_apps,
             app_obj = self.MARATHON_APP_TEMPLATES[app_type]
             app_obj = deepcopy(app_obj)
 
-            if num_instances == 0:
+            if num_instances:
+                if num_instances < 0:
+                    app_obj['instances'] = index % 3 + 1
+                else:
+                    app_obj['instances'] = num_instances
+            elif num_instances == 0:
```
grastogi23, line 194: you don't need elif... if num_instances would be false for 0, none etc.

PR #59, opened 2016-05-23
SDK: sample code for how to use AVI_API_TOKEN for authentication.

skdas:


README.md
```diff
@@ -73,6 +73,16 @@ for vs in resp.json()['results']:
...
 resp = api.delete('virtualservice', 'sample_vs')
 ```
+
+If ApiSession is invoked in the context of a control script, then token can be used for authentication.
+Along with that, information regarding username and tenant information can also be retrieved as follows: 
+```python
+token=os.environ.get('API_TOKEN')
+user=os.environ.get('USER')
+tenant=os.environ.get('TENANT')
+api = ApiSession.get_session("localhost", user, token=token, tenant=tenant)
```
grastogi23, line 83: nice...

PR #72, opened 2016-07-06
Av 10970 deb rpm pkgs

chaitanyaavi:


python/avi/sdk/setup.py
```diff
@@ -18,7 +18,7 @@
     url='http://avinetworks.com/',
     author='Avi Networks',
     author_email='support@avinetworks.com',
-    scripts=['avi/sdk/samples/virtualservice_examples_api.py'],
+    scripts=['samples/virtualservice_examples_api.py'],
```
grastogi23: this does not seem right... if we dont' do avi/sdk/ then it would not put it at avi/sdk/ directory

PR #74, opened 2016-07-11
Changes for F5 partitions and route domain 

chaitanyaavi:


python/avi/f5_converter/conversion_util.py
```diff
@@ -598,3 +624,27 @@ def create_header_rule(name, hdr_name, match, action, val, rule_index):
...
+        vrf_obj = {
+            "name": vrf_name,
+            "system_default": False
+        }
+        vrf_list.append(vrf_obj)
+
+
+def get_tenant_ref(name):
+    tenant = None
+    if name.startswith('/'):
```
grastogi23, line 643: can the ref be http:// instead of /api/xxx

chaitanyaavi, line 643: Here we are splitting the object name from F5 which is in the format /<partition-name>/<object-name> eg. /CTO/Atmos-QDC_8080 where CTO is partition name and Atmos-QDC_8080  is vs name we are ignoring /Common/ in parsing hence if the object is of common partition then it's name will not start with '/'

PR #77, opened 2016-07-13
The following are applicable for this changeset:

nmandeyam:
1. Rest-apis can use a different port other than the standard ports.
   Introduced keyword port=None to get_session and ApiSession init
   functions.
2. Rewired the get_full_path api to include the port-number in the
   path when 'port' is not none.  Otherwise, use the prefix.
3. When multiple sessions with the same username BUT different IP
   addresses are used, the library was only keying on the username.
   To enhance the library to setup sessions to different controllers
   with the same username, introduced a new key mechanism:
   key = controller_ip + username.
   
   Note: no impact from external API perspective.  The dict key is
   'controller_ip + username' instead of just username.
4. Introduced 'delete_session' to remove the cached entry irrespective
   of the activity-timer.  This api is useful in the credentials change
   context, where we have to clear the cached entry.
5. Whenever authentication fails, do a clear of the cached entry.

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -112,7 +112,7 @@ class ApiSession(Session):
     SESSION_CACHE_EXPIRY = 20*60
 
     def __init__(self, controller_ip, username, password=None, token=None,
-                 tenant=None, tenant_uuid=None, verify=False):
+                 tenant=None, tenant_uuid=None, verify=False, port=None):
         """
         initialize new session object with authenticated token from login api.
         It also keeps a cache of user sessions that are cleaned up if inactive
@@ -129,14 +129,16 @@ class ApiSession(Session):
         self.prefix = (controller_ip if controller_ip.startswith('http')
                        else "https://%s" % controller_ip)
         self.verify = verify
+        self.port = port
+        self.key = controller_ip + username
 
         try:
-            user_session = ApiSession.sessionDict[username]["api"]
+            user_session = ApiSession.sessionDict[self.key]["api"]
         except KeyError:
-            logger.debug("Session does not exist creating new session for %s",
-                         username)
+            logger.debug("Session does not exist; Creating new session for %s",
+                         self.key)
             self.authenticate_session()
-            ApiSession.sessionDict[username] = \
+            ApiSession.sessionDict[self.key] = \
                 {"api": self, "last_used": datetime.utcnow()}
             user_session = self
         self.headers = copy.deepcopy(user_session.headers)
@@ -152,7 +154,7 @@ class ApiSession(Session):
 
     @staticmethod
     def get_session(controller_ip, username, password=None, token=None,
-                    tenant=None, tenant_uuid=None, verify=False):
+                    tenant=None, tenant_uuid=None, verify=False, port=None):
         """
         returns the session object for same user and tenant
         calls init if session dose not exist and adds it to session cache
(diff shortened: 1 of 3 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -129,14 +129,16 @@ def __init__(self, controller_ip, username, password=None, token=None,
         self.prefix = (controller_ip if controller_ip.startswith('http')
                        else "https://%s" % controller_ip)
         self.verify = verify
+        self.port = port
+        self.key = controller_ip + username
```
anpartha: Cosmetic comment: May be a delimiter between username and controller ip. Perhaps username@controller_ip

python/avi/sdk/avi_api.py
```diff
@@ -112,7 +112,7 @@ class ApiSession(Session):
     SESSION_CACHE_EXPIRY = 20*60
 
     def __init__(self, controller_ip, username, password=None, token=None,
-                 tenant=None, tenant_uuid=None, verify=False):
+                 tenant=None, tenant_uuid=None, verify=False, port=None):
```
ypraveen, line 114: Ports other than 443 can already be supported by specifying controller_ip as the full path e.g., controller_ip=https://ip:9443/. It is not clean but it avoids yet another argument.

python/avi/sdk/avi_api.py
```diff
@@ -129,14 +129,16 @@ def __init__(self, controller_ip, username, password=None, token=None,
         self.prefix = (controller_ip if controller_ip.startswith('http')
```
ypraveen, line 128: If we decide to keep the port as separate option, then update the self.prefix string with port right here rather than doing it during construction of api_path.

python/avi/sdk/avi_api.py
```diff
@@ -535,9 +551,19 @@ def _clean_inactive_sessions():
...
-            logger.debug("Removed session for : %s", user)
-            del session_cache[user]
+            logger.debug("Removed session for : %s", key)
+            del session_cache[key]
+
+    def delete_session(self):
+        """ Removes the session for cleanup"""
+        logger.debug("Removed session for : %s", self.key)
+        try:
+            ApiSession.sessionDict[self.key]
```
ypraveen: Don't you need to do pop or del to delete?

python/avi/sdk/avi_api.py
nmandeyam: Accept. I will fix it.

python/avi/sdk/avi_api.py
nmandeyam, line 114: The Referer has to be ONLY the IP address.  It cannot have IPaddress:port for CSRF purposes.
For login purposes, it has be the IP address, while for post, it has to be IP:port.  Thus the need
to keep prefix separate and conjoin it port for full-path.

python/avi/sdk/avi_api.py
nmandeyam, line 128: Please review previous comment on this front.  port cannot be embedded with IP for CSRF purposes.

python/avi/sdk/avi_api.py
nmandeyam: Good catch.  I think i missed it while stashing it.

python/avi/sdk/avi_api.py
```diff
@@ -535,9 +550,19 @@ def _clean_inactive_sessions():
...
-            logger.debug("Removed session for : %s", user)
-            del session_cache[user]
+            logger.debug("Removed session for : %s", key)
+            del session_cache[key]
+
+    def delete_session(self):
+        """ Removes the session for cleanup"""
+        logger.debug("Removed session for : %s", self.key)
+        try:
+            ApiSession.sessionDict.pop(self.key)
```
grastogi23: you can call ApiSession.sessionDict.pop(self.key, None) instead

PR #79, opened 2016-07-15
Fixed bug in the ApiSession factory where it cannot recover when user…

grastogi23:
… name-password or token

changes in the server and caller is trying to use new credentials.

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -168,15 +168,20 @@ class ApiSession(Session):
         key = controller_ip + ":" + username
         try:
             user_session = ApiSession.sessionDict[key]["api"]
-            if user_session.password != password:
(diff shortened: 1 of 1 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -168,15 +168,19 @@ def get_session(controller_ip, username, password=None, token=None,
         key = controller_ip + ":" + username
         try:
             user_session = ApiSession.sessionDict[key]["api"]
-            if user_session.password != password:
-                logger.debug('Password mismatch; Clean cached session %s', key)
+            tenant = tenant if tenant else 'admin'
+            if (user_session.password != password or
+                user_session.keystone_token != token or
+                user_session.tenant != tenant):
```
ypraveen: check tenant_uuid mismatch too?

grastogi23: updated to check for tenant

PR #83, opened 2016-07-25
Added support for tcp half open and other comment fixes

chaitanyaavi:
Added support for tcp half open and other comment fixes


grastogi23:
again, we need to add more unit tests for all these. let us create a task for doing it.

PR #85, opened 2016-08-03
helper functions: Return none when input obj is none

ypraveen:
For functions get_obj_uuid and get_obj_ref


ypraveen:
please check the second commit that fixes a bug we have observed in heat integration.

PR #86, opened 2016-08-04
Av 8923

chaitanyaavi:


python/avi/netscaler_converter/lbvs_converter.py
```diff
@@ -50,22 +55,60 @@ def convert(self, ns_config, avi_config):
...
+                if ntwk_prof:
+                    vs_obj['network_profile_ref'] = ntwk_prof
                 avi_config['VirtualService'].append(vs_obj)
                 conv_status = ns_util.get_conv_status(
                     lb_vs, self.skip_attrs, self.na_attrs, self.indirect_list)
                 ns_util.add_conv_status(cmd, conv_status, vs_obj)
+
+                if enable_ssl:
+                    ssl_mappings = ns_config.get('bind ssl vserver', {})
+                    for mapping in ssl_mappings.get(key, []):
```
grastogi23, line 99: if mapping is an array then it should not be mapping.keys(). If it is a a dict then you can do maping.keys(). why not just do 
if 'CA' in mapping:

python/avi/netscaler_converter/test/test_monitor_converter.py
```diff
@@ -0,0 +1,87 @@
...
+    gSAMPLE_CONFIG = json.loads(cfg)
+    log.debug(' read cofig %s', gSAMPLE_CONFIG)
+    status_file = "./test_output" + os.path.sep + "ConversionStatus.csv"
+    csv_file = open(status_file, 'w')
+    ns_util.add_csv_headers(csv_file)
+
+
+class Test(unittest.TestCase):
+
+    LOG = logging.getLogger("converter-log")
```
grastogi23: why logging is within the class. you can just setup one logger for the whole test file.

python/avi/netscaler_converter/test/test_ns_service_converter.py
```diff
@@ -0,0 +1,43 @@
...
+    fh.setFormatter(formatter)
+    LOG.addHandler(fh)
+
+    def test_service_conversion(self):
+        service_converter = ServiceConverter()
+        avi_config = gSAMPLE_CONFIG["avi_config"]
+        ns_config_dict = gSAMPLE_CONFIG["ns_config_dict"]
+        service_converter.convert(ns_config_dict, avi_config)
+        assert avi_config['Pool'][0]
+        assert len(avi_config['Pool'][0]['health_monitor_refs']) == 3
```
grastogi23: is it possible to tie it to something in the ns_config_dict like len(pool) etc.

chaitanyaavi: Yes we can do that but its not straightforward we have to do conversion like code as it is generated from  bindings of services and service groups which are associated to same VS.

python/avi/netscaler_converter/lbvs_converter.py
chaitanyaavi, line 99: Done

grastogi23, line 99: I don't see it updated in the pull

PR #90, opened 2016-08-12
Added support for OnConnect profile

chaitanyaavi:
Updated represntation for indirect mapping in CSV file


grastogi23:
Please add unit test for the one connect.

PR #91, opened 2016-08-12
Fixed the bug where delete wouldn't accept data. This is needed for m…

grastogi23:
…acro api


python/avi/sdk/avi_api.py
```diff
@@ -467,7 +467,7 @@ def delete(self, path, tenant='', tenant_uuid='', timeout=60, params=None,
             parameters
```
ypraveen, line 467: doc for this new argument?

PR #94, opened 2016-08-17
added better check for the reference fields

grastogi23:


python/avi/sdk/utils/api_utils.py
```diff
@@ -32,6 +33,8 @@ def import_ssl_certificate(self, name, key, cert, key_passphrase='',
                 tenant=tenant, tenant_uuid=tenant_uuid)
         return resp
 
+RE_REF_MATCH = re.compile('/api/[\w/]+\?name\=[\w#]+')
```
ypraveen: As far as I know we don't allow '#' in our names, why match that in name?

python/avi/sdk/utils/api_utils.py
```diff
@@ -42,19 +45,18 @@ def ref_n_str_cmp(x, y):
...
         parts = uuid_or_name.split('#')
         y_uuid = parts[0]
-        y_name = parts[1] if len(parts) > 1 else ''
-        result = (x == y_uuid) or (x == y_name)
-        return result
-    return (x == y)
+        y_name = parts[1] if len(parts) > 1 else y_uuid
+    if RE_REF_MATCH.match(x):
+        x = x.split('name=')[1]
+    return (x == y_uuid) or (x == y_name)
```
ypraveen, line 60: we should probably split off anything after "#"?

grastogi23, line 60: not required. from webapp we don't add anything other than name after the # this is not a general url string but reference returned from the django. I am pushing updated re to exclude the #

PR #95, opened 2016-08-17
Added test for complete conversion

chaitanyaavi:
Added fixes for erros observed in Fandango and Apple config conversion

### DIFF
--- a/python/avi/netscaler_converter/ns_service_converter.py
+++ b/python/avi/netscaler_converter/ns_service_converter.py
@@ -17,6 +17,8 @@ class ServiceConverter(object):
 
     server_skip = ['comment', 'td']
 
+    skip_for_val = {'useproxyport': 'YES', 'maxReq': '0'}
+
     def convert(self, ns_config, avi_config):
         """
         Converts service or service groups bound to VS to avi Pool entity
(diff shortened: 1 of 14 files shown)


python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -80,7 +82,7 @@ def get_servers(self, ns_config, group, conv_status):
             ns_service = None
             if len(member.get('attrs', [])) > 1:
                 ns_service = ns_services.get(member['attrs'][1], None)
-            if not ns_service:
+            if not ns_service and len(member['attrs']) > 1:
```
grastogi23, line 85: are you checking for > 1 or > 0?. Can you pl. explain the code here.

chaitanyaavi, line 85: In command
bind lb vserver <name>@ ((<serviceName>@ [-weight <positive_integer>] ) | <serviceGroupName>@ | (-policyName <string>@ [-priority <positive_integer>] [-gotoPriorityExpression <expression>] [-type ( REQUEST | RESPONSE )] [-invoke (<labelType> <labelName>) ] ))
If it is bound to service or service group it has 2 attributes and other optional params but if it is bound to policy it has only one attribute and optional params so this condition checks if this bind command is for policy or service/servicegroup

PR #104, opened 2016-08-26
Sdk gr

grastogi23:


ypraveen:
Ready for review?
Also, this looks like a good candidate for squash-n-merge (all commits merged into one).. Let me know if you want me to keep them separate.

grastogi23:
other than metrics and health score everything is related to ansible... so we can do a squash-n-merge..is there way to do it from github or I update the commit?

ypraveen:
I can do it in github.

PR #107, opened 2016-09-01
F5 converter

chaitanyaavi:


python/avi/f5_converter/monitor_converter.py
```diff
@@ -550,26 +514,21 @@ def convert_udp(self, monitor_dict, f5_monitor, skipped):
...
-                udp_monitor["maintenance_response"] = maintenance_response
-            else:
-                udp_monitor = {"maintenance_response": maintenance_response}
-                monitor_dict["udp_monitor"] = udp_monitor
+        maintenance_resp = self.get_maintenance_response(f5_monitor)
+        if udp_monitor:
+            udp_monitor["maintenance_response"] = maintenance_resp
+        else:
+            udp_monitor = {"maintenance_response": maintenance_resp}
+            monitor_dict["udp_monitor"] = udp_monitor
```
grastogi23: shouldn't this be out of the else. Eg. in case of the if udp_monitor it would not update the monitor_dict

python/avi/f5_converter/monitor_converter.py
```diff
@@ -607,4 +566,19 @@ def convert_external(self, monitor_dict, f5_monitor, skipped,
             "command_variables": script_vars
         }
         monitor_dict["external_monitor"] = ext_monitor
-        return skipped
\ No newline at end of file
+        return skipped
+
+    def get_maintenance_response(self, f5_monitor):
+        maintenance_response = ''
+        if "reverse" in f5_monitor.keys():
```
grastogi23: you don't need f5_monitor.keys(). Just do if "reverse" in f5_monitor

python/avi/f5_converter/monitor_converter.py
```diff
@@ -607,4 +566,19 @@ def convert_external(self, monitor_dict, f5_monitor, skipped,
...
+            maintenance_response = f5_monitor.get("recv", '')
+        elif "recv disable" in f5_monitor.keys():
```
grastogi23: same here no check for .keys()

python/avi/f5_converter/monitor_converter.py
```diff
@@ -607,4 +566,19 @@ def convert_external(self, monitor_dict, f5_monitor, skipped,
...
+            maintenance_response = f5_monitor.get("recv disable", '')
```
grastogi23: here it is guarantee that f5_monitor has "recv_disable" so simply maintenance_response = f5_monitor['recv disable']

python/avi/f5_converter/monitor_converter.py
```diff
@@ -607,4 +566,19 @@ def convert_external(self, monitor_dict, f5_monitor, skipped,
             "command_variables": script_vars
         }
         monitor_dict["external_monitor"] = ext_monitor
-        return skipped
\ No newline at end of file
+        return skipped
+
+    def get_maintenance_response(self, f5_monitor):
```
grastogi23, line 576: add doc string

python/avi/f5_converter/monitor_converter.py
```diff
@@ -607,4 +566,19 @@ def convert_external(self, monitor_dict, f5_monitor, skipped,
...
+        maintenance_response = ''
+        if "reverse" in f5_monitor.keys():
+            maintenance_response = f5_monitor.get("recv", '')
+        elif "recv disable" in f5_monitor.keys():
+            maintenance_response = f5_monitor.get("recv disable", '')
+        if maintenance_response:
+            maintenance_response = \
+                maintenance_response.replace('\"', '').strip()
+        if maintenance_response == 'none':
```
grastogi23, line 590: Just checking if it is 'none' or None

chaitanyaavi: Its none this comes from F5 config which has value as 'none'

PR #113, opened 2016-09-15
Sample config backup script

khaltore:
@anpartha I'm committing your script to samples directory in sdk

### DIFF
--- a/python/avi/sdk/samples/avi_config_backup.py
+++ b/python/avi/sdk/samples/avi_config_backup.py
@@ -1,0 +1,37 @@
+import json
+import os
+import requests
+import sys
+import argparse
+import time
+
+from avi.sdk.avi_api import ApiSession
+
+def backup_configuration(api, args):
+    data = {'passphrase': args.passphrase} if args.passphrase else {}
+    params = {'full_system': True}
+    filename = "%s_%s.%s" % (args.filename, args.controller, time.strftime("%Y%m%d_%H%M%S"))
+    rsp = api.post('configuration/export', params=params, data = json.dumps(data))
+    with open(filename, 'w') as f:
+        f.write(rsp.text)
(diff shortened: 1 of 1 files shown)


python/avi/sdk/samples/avi_config_backup.py
```diff
@@ -0,0 +1,57 @@
...
+import argparse
+import time
+
+def api_uri(args, path):
+    return os.path.join("https://", args.controller, path)
+
+def login(session, args):
+    data = {'username': args.username, 'password': args.password}
+    headers = {'Content-Type': 'application/json'}
+    rsp = session.post(api_uri(args, "login"), verify=False, headers=headers,
```
grastogi23: Let us convert this into SDK rather than normal session. Users will shoot themselves if they don't handle CSRF token etc. also, it would shrink the code significantly

grastogi23 (changes requested):
Pl. change this to use avisdk.

khaltore:
I have changed it to use the sdk session. The only thinng is that they need to pip install the sdk. But that is fine.

python/avi/sdk/samples/avi_config_backup.py
```diff
@@ -0,0 +1,37 @@
...
+import argparse
+import time
+
+from avi.sdk.avi_api import ApiSession
+
+def backup_configuration(api, args):
+    data = {'passphrase': args.passphrase} if args.passphrase else {}
+    params = {'full_system': True}
+    filename = "%s_%s.%s" % (args.filename, args.controller, time.strftime("%Y%m%d_%H%M%S"))
+    rsp = api.post('configuration/export', params=params, data = json.dumps(data))
```
grastogi23, line 14: generally don't want to have spaces in key, value args as data=json.dumps(data).

grastogi23 (approved):

PR #118, opened 2016-09-21
Referer API fixes

nmandeyam:
# Summary:
1. If mode is https and port is none or 443, we don't embed the
   port in the prefix. The prefix would be 'https://ip'. If port
   is a non-default value then we concatenate https://ip:port
   in the prefix.
2. If mode is http and the port is none or 80, we don't embed the
   port in the prefix. The prefix would be 'http://ip'. If port is
   a non-default value, then we concatenate http://ip:port in
   the prefix.
# UT/IT:
1. UT tests.
2. gslb_gsvc run
   03.ITed on 2 controllers (GSLB)

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -117,6 +117,16 @@ class ApiSession(Session):
         initialize new session object with authenticated token from login api.
         It also keeps a cache of user sessions that are cleaned up if inactive
         for more than 20 mins.
+
+        Notes:
+        01. If mode is https and port is none or 443, we don't embed the
+            port in the prefix. The prefix would be 'https://ip'. If port
+            is a non-default value then we concatenate https://ip:port
+            in the prefix.
+        02. If mode is http and the port is none or 80, we don't embed the
+            port in the prefix. The prefix would be 'http://ip'. If port is
+            a non-default value, then we concatenate http://ip:port in
+            the prefix.
         """
         super(ApiSession, self).__init__()
         self.controller_ip = controller_ip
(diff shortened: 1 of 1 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -126,12 +136,21 @@ def __init__(self, controller_ip, username, password=None, token=None,
         self.tenant_uuid = tenant_uuid
         self.tenant = tenant if tenant else "admin"
         self.headers = {}
-        self.prefix = (controller_ip if controller_ip.startswith('http')
-                       else "https://%s" % controller_ip)
         self.verify = verify
         self.port = port
         self.key = controller_ip + ":" + username
 
+        # Refer Notes 01 and 02
```
grastogi23, line 143: we should only check for the case when controller_ip does not start with http. when user is specifying the controller uri then port should also be provided.

python/avi/sdk/avi_api.py
```diff
@@ -126,12 +136,21 @@ def __init__(self, controller_ip, username, password=None, token=None,
...
+        if controller_ip.startswith('http'):
+            if port is None or port == 80:
+                self.prefix = controller_ip
+            else:
+                self.prefix = '{x}:{y}'.format(x=controller_ip, y=port)
+        else:
+            if port is None or port == 443:
```
grastogi23, line 150: the check would simply be if port in (80, 443, None) then ignore.

grastogi23 (changes requested):

grastogi23 (approved):

PR #121, opened 2016-09-28
Netscaler converter : Fix for VS State should be disabled by default

chaitanyaavi:


python/avi/netscaler_converter/lbvs_converter.py
```diff
@@ -44,9 +44,12 @@ def convert(self, ns_config, avi_config):
                 ip_addr = lb_vs['attrs'][2]
                 port = lb_vs['attrs'][3]
 
-                state = lb_vs.get('state', 'ENABLED')
-                enabled = True
-                if state == 'DISABLED':
+                if vs_state == 'enable':
+                    state = lb_vs.get('state', 'ENABLED')
```
grastogi23: you can change this to be enabled = (lb_vs.get('state', 'ENABLED') == 'ENABLED')

grastogi23 (changes requested):

python/avi/netscaler_converter/lbvs_converter.py
chaitanyaavi: Done

PR #144, opened 2016-11-21
Fixed hm count

rameshtathe:
I have made some changes, please review
1) Updated csvs_converter.py for HTTP.REQ.COOKIE.CONTAINS and equal rules.
2) Updated ns_service_converter.py of hm ref in pool for maximum count of HM

### DIFF
--- a/python/avi/netscaler_converter/ns_service_converter.py
+++ b/python/avi/netscaler_converter/ns_service_converter.py
@@ -61,7 +61,7 @@ class ServiceConverter(object):
 
                 monitor_refs = [mon["name"] for mon in monitors]
 
-                pool_obj["health_monitor_refs"] = monitor_refs
+                pool_obj["health_monitor_refs"] = list(set(monitor_refs))
                 avi_config['Pool'].append(pool_obj)
 
                 for status in conv_status:
--- a/python/avi/netscaler_converter/ns_service_converter.py
+++ b/python/avi/netscaler_converter/ns_service_converter.py
@@ -50,15 +50,16 @@ class ServiceConverter(object):
                     }
                 monitor_names = self.get_monitors(ns_config, group)
                 avi_monitors = avi_config["HealthMonitor"]
-                monitor_refs = []
+                hm_monitors = []
                 for ref in monitor_names:
                     refs = [mon for mon in avi_monitors
                             if mon['name'] == ref]
                     if refs:
-                        monitor_refs.append(ref)
-                    else:
-                        LOG.warn('Health monitor %s not found in avi config'
-                                 % ref)
+                        hm_monitors.append(refs[0])
+
+                monitors = self.remove_duplicate_health_monitors(hm_monitors)
+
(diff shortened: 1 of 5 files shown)


python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -189,3 +190,24 @@ def convert_ns_service_group(self, ns_service_group, ns_servers,
             }
             servers.append(server_obj)
         return servers
+
+    def remove_duplicate_health_monitors(self, monitors):
+        if len(monitors) == 1:
```
chaitanyaavi: The logic should be generic it will fail for new supported attribute added check for duplicate policy remove logic in conversion_util.py of F5 converter for example.

chaitanyaavi (changes requested):

python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -50,15 +50,16 @@ def convert(self, ns_config, avi_config):
...
-                    else:
-                        LOG.warn('Health monitor %s not found in avi config'
-                                 % ref)
+                        hm_monitors.append(refs[0])
+
+                monitors = self.remove_duplicate_health_monitors(hm_monitors)
+
+                monitor_refs = [mon["name"] for mon in monitors]
 
                 pool_obj["health_monitor_refs"] = list(set(monitor_refs))
```
grastogi23, line 64: no need to do list(set(monitor_refs) as monitor_refs is already going to be a list.

python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -189,3 +190,24 @@ def convert_ns_service_group(self, ns_service_group, ns_servers,
             }
             servers.append(server_obj)
         return servers
+
+    def remove_duplicate_health_monitors(self, monitors):
```
grastogi23: add method documentation

grastogi23:
Pl. also add test case for these changes.

grastogi23 (changes requested):

grastogi23:
Also, would like these fixes to be merged to 16.3 release branch as well.

grastogi23 (approved):

grastogi23:
waiting on chaitanya to approve the request

python/avi/netscaler_converter/ns_service_converter.py
chaitanyaavi: It should make copy of object remove name and description from that and do dict comparison for duplicates
If we support any other attribute other than this condition in HM conversion this condition also need to be modified

python/avi/netscaler_converter/ns_service_converter.py
chaitanyaavi: This is done to remove duplicates, set will remove the duplicates and for assigning to refs converted again to list

chaitanyaavi (approved):

PR #148, opened 2016-11-24
Added test cases for remove duplicate health monitors

rameshtathe:
I have made changes, Please review
1) Updated test_ns_service_converter.cfg
2) test_ns_service_converter.py

### DIFF
--- a/python/avi/netscaler_converter/test/test_ns_service_converter.py
+++ b/python/avi/netscaler_converter/test/test_ns_service_converter.py
@@ -100,3 +100,12 @@ class Test(unittest.TestCase):
                 else:
                     count += 1
             return count
+
+    def remove_duplicate_health_monitors(self):
+        converted_hm_config = gSAMPLE_CONFIG["converted_hm_config"]
+        hm_config = gSAMPLE_CONFIG["hm_config"]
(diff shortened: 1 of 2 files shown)


chaitanyaavi (changes requested):
I dont think we need this granular test cases for util methods add the scenario in the input data to cover this code and add assert for this condition.
@grastogi23 What are your thoughts on this?

grastogi23:
I think it is good to have unit test case for the util methods. However, we should also add tests for covering some of the this code path in top level scenarios.

python/avi/netscaler_converter/test/test_ns_service_converter.py
```diff
@@ -99,4 +99,13 @@ def get_server_count(self, ns_config):
                             count += 1
                 else:
                     count += 1
-            return count
\ No newline at end of file
+            return count
+
+    def remove_duplicate_health_monitors(self):
```
grastogi23, line 104: also, this is not a test... which code will call this function for testing?

grastogi23 (changes requested):

grastogi23 (approved):

PR #158, opened 2016-12-09
AV-15619: Cherry-pick to Master

nmandeyam:
Summary:
=======
01. AV-15619:
    -> SDK layer does not  pass the timeout in avi_api(s)
    -> Introduced timeout parameter in the initialization and get_session
       functions.
    -> If timeout is NOT passed, then it will default to 60 seconds
    -> Save the timeout value in the session block.
    -> Use the timeout value in a consistent manner for the CRUD APIs.
    -> Reorder imports to avoid pylint errors.

======
01. UT/IT:

------------------------------------------------------------------------------
gslb site.txt WITH gslb site a.conf,gslb site b.conf,gslb site glo... | PASS |
14 critical tests, 14 passed, 0 failed
14 tests total, 14 passed, 0 failed
==============================================================================

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -128,8 +128,6 @@ class ApiSession(Session):
             port in the prefix. The prefix would be 'http://ip'. If port is
             a non-default value, then we concatenate http://ip:port in
             the prefix.
-        03. If the timeout value is None or zero, then default the value to 60. 
-            Otherwise, use the passed value.
         """
         super(ApiSession, self).__init__()
         self.controller_ip = controller_ip
@@ -154,11 +152,7 @@ class ApiSession(Session):
                 self.prefix = 'https://{x}'.format(x=controller_ip)
             else:
                 self.prefix = 'https://{x}:{y}'.format(x=controller_ip, y=port)
-
-        # Refer Notes 03
         self.timeout = timeout
-        if timeout is None or timeout==0:
-            self.timeout = 60
         try:
             user_session = ApiSession.sessionDict[self.key]["api"]
         except KeyError:
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -113,7 +113,7 @@ class ApiSession(Session):
 
     def __init__(self, controller_ip, username, password=None, token=None,
                  tenant=None, tenant_uuid=None, verify=False, port=None,
-                 timeout=None):
+                 timeout=60):
         """
         initialize new session object with authenticated token from login api.
         It also keeps a cache of user sessions that are cleaned up if inactive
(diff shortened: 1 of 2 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -127,6 +128,8 @@ def __init__(self, controller_ip, username, password=None, token=None,
             port in the prefix. The prefix would be 'http://ip'. If port is
             a non-default value, then we concatenate http://ip:port in
             the prefix.
+        03. If the timeout value is None, then default the value to 60. 
+            Otherwise, use the passed value.
```
ypraveen: Curious: why not set timeout=60 in function arg definition?

nmandeyam: Accept: I will address it on master.

python/avi/sdk/avi_api.py
```diff
@@ -151,6 +154,11 @@ def __init__(self, controller_ip, username, password=None, token=None,
                 self.prefix = 'https://{x}'.format(x=controller_ip)
             else:
                 self.prefix = 'https://{x}:{y}'.format(x=controller_ip, y=port)
+
+        # Refer Notes 03
+        self.timeout = timeout
+        if timeout is None or timeout==0:
```
ypraveen: I am assuming that you are trying to catch the cases when someone passes bad argument to timeout. In that case, we should check if the passed value for timeout is a positive integer or not.

If that is not the goal, we can skip this check.

python/avi/sdk/avi_api.py
```diff
@@ -279,7 +289,7 @@ def _get_api_headers(self, tenant, tenant_uuid, timeout, headers):
         return api_hdrs
 
     def _api(self, api_name, path, tenant, tenant_uuid, data=None,
-             headers=None, timeout=60, **kwargs):
+             headers=None, timeout=None, **kwargs):
```
ypraveen, line 286: I think we don't need to make this change, and also at several other places down below.

python/avi/sdk/avi_api.py
```diff
@@ -317,15 +330,15 @@ def _api(self, api_name, path, tenant, tenant_uuid, data=None,
...
+                             headers=headers, timeout=timeout, **kwargs)
             self.num_session_retries = 0
         if resp.cookies and 'csrftoken' in resp.cookies:
             csrftoken = resp.cookies['csrftoken']
             self.headers.update({"X-CSRFToken": csrftoken})
         self._update_session_last_used()
         return ApiResponse.to_avi_response(resp)
 
-    def get(self, path, tenant='', tenant_uuid='', timeout=60, params=None,
+    def get(self, path, tenant='', tenant_uuid='', timeout=None, params=None,
```
ypraveen, line 335: same here.

grastogi23: these are needed

python/avi/sdk/avi_api.py
grastogi23: these are needed

python/avi/sdk/avi_api.py
grastogi23: agree with @ypraveen we don't need this check

grastogi23 (changes requested):

python/avi/sdk/avi_api.py
grastogi23: needed to allow individual methods to override what was set in the constructor

grastogi23 (approved):

PR #161, opened 2016-12-13
16.3 ns changes

rameshtathe:


python/avi/netscaler_converter/lbvs_converter.py
```diff
@@ -85,11 +85,21 @@ def convert(self, ns_config, avi_config, vs_state):
...
-                    'pool_ref': pool_ref
                     }
 
+                if ip_addr == "0.0.0.0":
+                    ns_util.add_status_row(cmd, 'skipped')
+                    LOG.error("Skipped VS, Service point to %s server." % ip_addr)
+                    continue
+
+                service = {'port': port, 'enable_ssl': enable_ssl}
+                if port == "0" or port == "*":
```
grastogi23: also write as if port in ("0", "*")

python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -145,8 +152,16 @@ def convert_ns_service(self, ns_service, ns_servers, conv_status):
         if not state == 'ENABLED':
             enabled = False
         port = attrs[3]
-        if port == "*":
-            port = "0"
+        if port == "*" or port == "0":
+            port = "1"
+
```
grastogi23, line 158: should the port range also be set to be 65535.

python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -155,10 +170,14 @@ def convert_ns_service(self, ns_service, ns_servers, conv_status):
             'port': port,
             'enabled': enabled
         }
+        if not matches:
+            LOG.warning('Not found IP of server : %s' % ip_addr)
```
grastogi23: would it result in skip or just warning?

python/avi/netscaler_converter/ns_service_converter.py
```diff
@@ -193,5 +220,9 @@ def convert_ns_service_group(self, ns_service_group, ns_servers,
                 'port': port,
                 'enabled': enabled
             }
-            servers.append(server_obj)
+            if not matches:
+                LOG.warning('Not found IP of server : %s' % ip_addr)
```
grastogi23: should say ignoring the server and print the full ns server object.

python/avi/netscaler_converter/ns_util.py
```diff
@@ -197,4 +197,24 @@ def remove_duplicate_objects(obj_type, obj_list):
...
+    return obj_list
+
+def remove_duplicate_server_objects(obj_type, obj_list):
+    """
+    Remove duplicate objects from list
+    :param obj_type: Object type
+    :param obj_list: list of all objects
+    :return: return list which has no duplicates objects
+    """
+    for source_obj in obj_list:
```
grastogi23: this code needs re-write. If you are trying to find duplicates then simply put this in a hash and compare the objects or just put it in a set and then make list of out.

grastogi23 (changes requested):

python/avi/netscaler_converter/ns_util.py
```diff
@@ -197,4 +197,24 @@ def remove_duplicate_objects(obj_type, obj_list):
                 del obj_list[index]
                 remove_duplicate_objects(obj_type, obj_list)
 
-    return obj_list
\ No newline at end of file
+    return obj_list
+
+def remove_duplicate_server_objects(obj_type, obj_list):
```
grastogi23: can you also write a unit test for the duplicates.

python/avi/netscaler_converter/ns_service_converter.py
rameshtathe: As we can see in VS proto, services has port_range_end but Servers does not have the port range in Pool proto
Please correct me if i am wrong

python/avi/netscaler_converter/ns_service_converter.py
rameshtathe: If you see, It return the list of servers. This function is call from get_servers and get servers called from convert. If it return the empty list then we skipped the bind lb vserver.

if not servers:
       ns_util.add_status_row(b_cmd, "Skipped")
       LOG.error('Error: No Servers found. Skipped pool : %s' % group_key)
       continue

python/avi/netscaler_converter/ns_util.py
rameshtathe: Fixed

grastogi23 (approved):

PR #168, opened 2017-01-02
New changes for Master

rameshtathe:
I have made following changes,

Log the build number for F5 and netscalar converter
Ignore more than 32 snat pools for F5 converter
Flag for ignoring default config for F5 converter
Support to convert _sys_https_redirect irule for F5 converter
Dummy datascript as placeholder for F5 Converter
Dummy ssl cert and key as placeholder for F5 converter


python/avi/f5_converter/f5_converter.py
```diff
@@ -6,7 +6,7 @@
 import sys
 
 from requests.packages import urllib3
-
+from avi.version import AVI_VERSION, AVI_PIP_VERSION
```
grastogi23: this would require that we package version in the controller too. we need to check if that is happening

python/avi/f5_converter/f5_converter.py
```diff
@@ -46,7 +47,10 @@ def get_default_config(version, is_download, path):
             monitor_dict["monitor"] = default_mon
             del monitor_dict["monitorroot"]
         profile_dict.update(monitor_dict)
-        f5_defaults_dict = profile_dict
+        if not skip_default_file:
+            f5_defaults_dict = profile_dict
+        LOG.warning('Skipped default file : %s' % profile_base.name)
```
grastogi23: change message to "Could not find default file: %s. Skipping and using converter defaults"

python/avi/f5_converter/f5_converter.py
```diff
@@ -59,14 +63,19 @@ def get_default_config(version, is_download, path):
...
+        LOG.warning('Skipped default file : %s' % defaults_file.name)
+
     return f5_defaults_dict
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument('-f', '--bigip_config_file',
                         help='absolute path for F5 config file')
+    parser.add_argument('--skip_default_file',
```
grastogi23: spelling. We don't need flag for this. it should be done automatically...

python/avi/f5_converter/f5_converter.py
```diff
@@ -130,6 +140,8 @@ def get_default_config(version, is_download, path):
         user_ignore = json.loads(ignore_conf_str)
 
     partitions = []
+    LOG.info('Avi Build version : %s' % AVI_VERSION)
```
grastogi23: Let us print this in the help string as well.

python/avi/f5_converter/vs_converter.py
```diff
@@ -169,8 +171,19 @@ def convert_vs(self, vs_name, f5_vs, vs_state, avi_config, snat_config,
             'enabled': enabled,
             'services': services_obj,
             'application_profile_ref': app_prof[0],
+            'vs_datascripts': []
         }
 
+        if 'rules' in f5_vs.keys():
```
grastogi23: check can just be if 'rules' in f5_vs

python/avi/f5_converter/vs_converter.py
```diff
@@ -218,7 +231,9 @@ def convert_vs(self, vs_name, f5_vs, vs_state, avi_config, snat_config,
             snat_pool = snat_config.pop(snat_pool_name, None)
         if snat_pool:
             snat_list = conv_utils.get_snat_list_for_vs(snat_pool)
-            vs_obj["snat_ip"] = snat_list
+            if len(snat_list):
```
grastogi23: let us re-write this logic.

If snat_list checks for len and is more efficient.
Print the logs when if len(snat_list) > 32 and log.

python/avi/f5_converter/vs_converter.py
```diff
@@ -264,6 +279,34 @@ def convert_vs(self, vs_name, f5_vs, vs_state, avi_config, snat_config,
 
         return vs_obj
 
+    def create_vs_datascript(self, rule, avi_config):
+        vs_ds_ref = rule + '-vs-datascript-dumy'
```
grastogi23: fix spelling dummy.

grastogi23 (changes requested):

python/avi/f5_converter/f5_converter.py
```diff
@@ -46,7 +47,10 @@ def get_default_config(version, is_download, path):
             monitor_dict["monitor"] = default_mon
             del monitor_dict["monitorroot"]
         profile_dict.update(monitor_dict)
-        f5_defaults_dict = profile_dict
+        if not skip_default_file:
```
chaitanyaavi: We need to have this check before parsing default config

python/avi/f5_converter/f5_converter.py
chaitanyaavi: No As discussed on the call for quick fix we need this flag right?

python/avi/f5_converter/profile_converter.py
```diff
@@ -114,9 +114,16 @@ def get_key_cert_obj(self, name, key_file_name, cert_file_name, input_dir,
         :param tenant: tenant name to add tenant ref in config
         :return:SSLKeyAndCertificate object
         """
-        folder_path = input_dir+os.path.sep
-        key = conv_utils.upload_file(folder_path + key_file_name)
-        cert = conv_utils.upload_file(folder_path + cert_file_name)
+        if not generate_ssl_placeholder:
```
chaitanyaavi: This should not be the flag driven, If the given cert file is not found on location then put the self cert and add warning in the log.

chaitanyaavi (changes requested):

PR #173, opened 2017-01-09
Fixed multiple ssl key and cert objects

rameshtathe:

1) Fixed multiple ssl key and cert objects.

### DIFF
--- a/python/avi/f5_converter/profile_converter.py
+++ b/python/avi/f5_converter/profile_converter.py
@@ -27,7 +27,7 @@ class ProfileConfigConv(object):
               'AES256-SHA:DES-CBC3-SHA'
 
     def convert_profile(self, profile, key, f5_config, profile_config,
-                        avi_config, input_dir, user_ignore, tenant_ref):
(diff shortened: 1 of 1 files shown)


python/avi/f5_converter/profile_converter.py
```diff
@@ -115,6 +117,13 @@ def get_key_cert_obj(self, name, key_file_name, cert_file_name, input_dir,
         :return:SSLKeyAndCertificate object
         """
 
+        cert_name = [cert['name'] for cert in key_and_cert_mapping_list if cert['key_file_name'] == key_file_name
+                     and cert['cert_file_name'] == cert_file_name]
+        if cert_name:
+            LOG.warning('SSL key and Certificate is already exist for %s and %s is %s'
+                        % (key_file_name, cert_file_name, cert_name[0]))
+            return None
```
grastogi23, line 125: shouldn't return be an object. if cert was already found?

grastogi23 (changes requested):

PR #177, opened 2017-01-11
api session: Saving the login response as it contains useful information

ypraveen:
login response contains the list of tenants that the logging user
has access to. This could be used to avoid making extra calls to
the Avi API server.

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -162,6 +162,7 @@ class ApiSession(Session):
         self.verify = verify
         self.port = port
         self.key = controller_ip + ":" + username
+        self.login_response = dict()
 
         # Refer Notes 01 and 02
         if controller_ip.startswith('http'):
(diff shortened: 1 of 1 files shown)


ypraveen:
I needed to find the tenant name given a tenant uuid of a user. However, a regular user can't make tenant APIs without proper privileges. Seems like login is the best way to get the information about tenants a user has access to.

ypraveen:
Looks like /api/user-tenant-list returns the tenant information of a user too. If you guys think it is not better to keep login response in ApiSession object, I am ok with that too.. just that I would need to make at least one more API call to get this info.

grastogi23:
I see problem of caching is when and how this gets updated? let us say when the API session was created user only had access to one tenant. later on it has access to two tenants. we need a way to indicate that this cache is now invalid. 

One more approach can be add a helper api/function in the ApiUtils. that way this can be a common code that can be easily shared but still has penalty of extra API call when it is needed by the caller.

ypraveen:
agreed; probably not really a good info to rely on.

PR #194, opened 2017-02-05
Changed csv report format

gauravopcito:

1) Made changes to update csv report fromat


python/avi/netscaler_converter/csvs_converter.py
```diff
@@ -212,9 +209,10 @@ def convert(self, ns_config, avi_config, vs_state):
                 conv_status = ns_util.get_conv_status(
                     bind_conf, self.bind_skipped, [], [])
                 if found:
-                    ns_util.add_conv_status(b_cmd, conv_status, None)
+
+                    ns_util.add_conv_status(b_cmd, vs_name, full_cmd, conv_status, vs_obj)
                 else:
-                    ns_util.add_status_row(b_cmd, 'skipped')
+                    ns_util.add_status_row(b_cmd, vs_name, full_cmd, 'Skipped')
```
grastogi23: it would be better to use a variable for 'skipped' rather than string so that we can replace these tags easily. Also, sometime skipped starts with small 's' and sometimes it starts with 'S'

python/avi/netscaler_converter/ns_util.py
```diff
@@ -177,16 +181,16 @@ def update_status_for_skipped(skipped_cmds):
...
                 break
         for id_cmd in indirect_cmds:
             if cmd.startswith(id_cmd):
-                add_status_row(cmd, 'Indirect')
+                add_status_row(id_cmd, '', cmd, 'Indirect')
                 is_id = True
                 break
         if not is_na and not is_id:
-            add_status_row(cmd, 'skipped')
+            add_status_row(cmd, '', '', 'skipped')
```
grastogi23: eg. here it is small 'skipped'

python/avi/netscaler_converter/ns_util.py
```diff
@@ -637,32 +640,36 @@ def policy_converter(policies, rule_index1, bind_patset, patset_config,
...
         if not is_policy_successful:
-            LOG.info('Datascript : %s' % bind_cmd)
-            add_status_row(bind_cmd, 'Datascript')
+            LOG.info('Datascript : %s %s' % (bind_cmd, b_cmd))
+            add_status_row(bind_cmd, b_cmd, full_cmd, 'Datascript')
         else:
-            LOG.info('Successful: %s' % bind_cmd)
-            add_status_row(bind_cmd, 'Successful')
+            LOG.info('Successful: %s %s' % (bind_cmd, b_cmd))
+            add_status_row(bind_cmd, b_cmd, full_cmd, 'Successful')
```
grastogi23: same here. 'Successful' vs 'successful'

python/avi/netscaler_converter/ns_util.py
```diff
@@ -872,7 +884,6 @@ def add_string_group_for_policy(string_group_name, matches, avi_config):
     for match in matches:
         stringgroup_object['kv'].append({"key": match})
     avi_config['StringGroup'].append(stringgroup_object)
-    add_status_row(string_group_name, "Successful")
```
grastogi23: seems like add_status_row was not added back?

grastogi23 (changes requested):

python/avi/netscaler_converter/ns_util.py
gauravopcito: It is not netscalar command, so removed from csv report.

python/avi/netscaler_converter/csvs_converter.py
```diff
@@ -56,17 +57,18 @@ def convert(self, ns_config, avi_config, vs_state):
...
             lbvs_bindings = []
             cs_vs = cs_vs_conf[key]
-            cmd = 'add cs vserver %s' % key
+            cmd = 'add cs vserver'
+            full_cmd = ns_util.get_netscalar_full_command(cmd, cs_vs)
             if not cs_vs['attrs'][1] in self.supported_types:
                 LOG.warn('Unsupported type %s of Context switch VS: %s' %
                          (cs_vs['attrs'][1], key))
-                ns_util.add_status_row(cmd, 'skipped')
+                ns_util.add_status_row(cmd, key, full_cmd, Status.SKIPPED.value)
```
grastogi23: Have you guys tested this. I should probably be Status.SKIPPED.name rather than Status.SKIPPED.value.

grastogi23 (changes requested):

PR #196, opened 2017-02-07
CSV/report changes

rameshtathe:

Made changes to update csv report fromat

### DIFF
--- a/python/avi/netscaler_converter/csvs_converter.py
+++ b/python/avi/netscaler_converter/csvs_converter.py
@@ -3,6 +3,7 @@ import copy
 import re
 from avi.netscaler_converter import ns_util
 from avi.netscaler_converter.lbvs_converter import Redirect_Pools
+from avi.netscaler_converter.ns_constants import STATUS_SKIPPED
 
 LOG = logging.getLogger(__name__)
 
@@ -56,17 +57,18 @@ class CsvsConverter(object):
             LOG.debug("Context Switch VS conversion started for: %s" % key)
             lbvs_bindings = []
             cs_vs = cs_vs_conf[key]
-            cmd = 'add cs vserver %s' % key
+            cmd = 'add cs vserver'
+            full_cmd = ns_util.get_netscalar_full_command(cmd, cs_vs)
             if not cs_vs['attrs'][1] in self.supported_types:
                 LOG.warn('Unsupported type %s of Context switch VS: %s' %
                          (cs_vs['attrs'][1], key))
-                ns_util.add_status_row(cmd, 'skipped')
+                ns_util.add_status_row(cmd, key, full_cmd, STATUS_SKIPPED)
                 continue
             tt = cs_vs.get('targetType', None)
             if tt and tt == 'GSLB':
                 LOG.warn('Unsupported target type %s of Context switch VS: %s' %
                          (cs_vs['attrs'][1], key))
-                ns_util.add_status_row(cmd, 'skipped')
+                ns_util.add_status_row(cmd, key, full_cmd, STATUS_SKIPPED)
             vs_name = cs_vs['attrs'][0]
             ip_addr = cs_vs['attrs'][2]
             port = cs_vs['attrs'][3]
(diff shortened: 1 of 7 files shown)


python/avi/netscaler_converter/ns_util.py
```diff
@@ -634,35 +641,42 @@ def policy_converter(policies, rule_index1, bind_patset, patset_config,
...
         if not is_policy_successful:
-            LOG.info('Datascript : %s' % bind_cmd)
-            add_status_row(bind_cmd, 'Datascript')
+            LOG.info('Datascript : %s %s' % (bind_cmd, b_cmd))
+            add_status_row(bind_cmd, b_cmd, full_cmd, STATUS_DATASCRIPT)
         else:
-            LOG.info('Successful: %s' % bind_cmd)
-            add_status_row(bind_cmd, 'Successful')
+            LOG.info('Successful: %s %s' % (bind_cmd, b_cmd))
+            add_status_row(bind_cmd, b_cmd, full_cmd, STATUS_SUCCESSFUL)
```
grastogi23: should there be an object passed down when status is successful?

python/avi/netscaler_converter/ns_util.py
```diff
@@ -689,7 +707,8 @@ def get_rewrite_action(policy_name, policy_rules, rewrite_action_config):
             if matches:
                 value = {'val': matches[0]}
                 policy_rule['hdr_action'][0]['hdr']['value'].update(value)
-                LOG.info('Successful: %s' % cmd)
+                LOG.info('Successful: %s %s' % (cmd, policy_name))
+                add_status_row(cmd, policy_name, full_cmd, STATUS_SUCCESSFUL)
```
grastogi23: same here. shouldn't we pass the policy object / rule here?

grastogi23:
can you pl. ensure all the conv_status with successful and partial have objects passed down.

python/avi/netscaler_converter/csvs_converter.py
```diff
@@ -157,12 +158,10 @@ def convert(self, ns_config, avi_config, vs_state):
                                 policy = copy.deepcopy(policy)
                                 policy['targetLBVserver'] = targetVserver
                                 cs_vs_policies.append(policy)
-                                ns_util.add_status_row(pl_cmd, 'successful')
-                                LOG.info('Successful : %s' % pl_cmd)
+                                LOG.info('Conversion successful : %s' % pl_cmd)
```
grastogi23, line 161: there is no add_status_row here. It seems to have been previously present

python/avi/netscaler_converter/csvs_converter.py
```diff
@@ -104,12 +106,11 @@ def convert(self, ns_config, avi_config, vs_state):
                 if clttimeout:
                     ns_util.add_clttimeout_for_http_profile(http_prof, avi_config, clttimeout)
                     clt_cmd = cmd + ' cltTimeout %s' % clttimeout
-                    ns_util.add_status_row(clt_cmd, 'Successful')
-                    LOG.info('Successful : %s' % clt_cmd)
+                    LOG.info('Conversion successful : %s' % clt_cmd)
```
grastogi23, line 109: I don't see replacement for add_status_row here.

grastogi23 (changes requested):

python/avi/netscaler_converter/csvs_converter.py
rameshtathe: @grastogi23 Here,  attribute clttimeout is successfully converted and added into the avi config object, though we add complete avi config object on completion of complete command conversion to CSV report some where below. So here it is not required to add_status_row.

python/avi/netscaler_converter/csvs_converter.py
rameshtathe: @grastogi23 Exactly same case as above. Here we have successfully converted attribute and not complete command yet so not added to CSV report but we only update AVI config object here which is logged to CSV report on completion of complete command.

grastogi23 (approved):

PR #213, opened 2017-02-28
Netscaler feature p2

rameshtathe:
I have made the following changes, Can you please review?
1.  Updated the yaml file with some new non applicable netscaler commands
2.  Updated the "Not in use" status to "Incomplete configuration"  
3. Moved all the ignore the value of priority expression if its equal to END


python/avi/netscaler_converter/ns_util.py
```diff
@@ -801,3 +801,14 @@ def create_http_policy_set_for_redirect_url(vs_obj, redirect_uri, avi_config, te
...
 
+
+def clean_virtual_service_from_avi_config(avi_config):
+    """
+    This function defines that clean up vs which has vip 0.0.0.0
+    :param avi_config: dict of AVI
+    :return: None
+    """
+    vs_list = copy.deepcopy(avi_config['VirtualService'])
+    avi_config['VirtualService'] = []
```
grastogi23, line 812: don't need to initialize to empty array as next state will do that always.

grastogi23 (approved):

PR #217, opened 2017-03-02
Fix for #AV-17472 : F5 V10 conversion issue

chaitanyaavi:


python/avi/f5_converter/vs_converter.py
```diff
@@ -193,7 +193,11 @@ def convert_vs(self, vs_name, f5_vs, vs_state, avi_config, snat_config,
         }
 
         if 'rules' in f5_vs:
-            rules = f5_vs['rules'].keys()
+            if isinstance(f5_vs['rules'], str):
```
grastogi23: I believe you will need to check for both basestring and unicode...

grastogi23 (changes requested):

python/avi/f5_converter/vs_converter.py
chaitanyaavi: Updated check for basestring which will check for str and unicode both.

grastogi23 (approved):

PR #220, opened 2017-03-16
Sprint-1 changes for 16.3

chaitanyaavi:


python/avi/f5_converter/__init__.py
```diff
@@ -1 +1 @@
-__version__ = ''
+__version__ = '16.3'
```
grastogi23: this shouldn't be done.. will cover in the meeting.

python/avi/f5_converter/f5_converter.py
```diff
@@ -118,8 +121,15 @@ def get_default_config(version, is_download, path, skip_default_file):
                         help='config json to skip the config in conversion')
     parser.add_argument('--partition_config',
                         help='comma separated partition config files')
+    parser.add_argument('--version',
+                        help='Print product version and exit',
+                        action='store_true')
 
     args = parser.parse_args()
+    # print avi f5 converter version
+    if args.version:
```
grastogi23, line 130: we should always print version when the tool is run and version is available. In case of the controller the version is not packaged so we should look for controller version if it exists and use that.

python/avi/f5_converter/f5_converter.py
```diff
@@ -144,8 +154,10 @@ def get_default_config(version, is_download, path, skip_default_file):
         user_ignore = json.loads(ignore_conf_str)
 
     partitions = []
-    #LOG.info('Avi Build version : %s' % AVI_VERSION)
-    #LOG.info('Avi pip version : %s' % AVI_PIP_VERSION)
+    # Add logger and print avi f5 converter version
+    LOG.info('AVI sdk version: %s' % __version__)
+    print 'AVI sdk version: %s' % __version__
```
grastogi23: print both f5 and sdk versions.

python/avi/f5_converter/f5_converter.py
```diff
@@ -144,8 +155,12 @@ def get_default_config(version, is_download, path, skip_default_file):
         user_ignore = json.loads(ignore_conf_str)
 
     partitions = []
-    #LOG.info('Avi Build version : %s' % AVI_VERSION)
-    #LOG.info('Avi pip version : %s' % AVI_PIP_VERSION)
+    # Add logger and print avi f5 converter version
+    LOG.info('AVI sdk version: %s Controller Version: %s'
+             % (__version__, args.controller_version))
+    print 'AVI sdk version: %s Controller Version: %s' \
+          % (__version__, args.controller_version)
```
grastogi23, line 162: we should change the default to be 16.3.8

python/avi/f5_converter/monitor_converter.py
```diff
@@ -122,8 +123,9 @@ def convert_monitor(self, f5_monitor, key, monitor_config, input_dir,
         description = f5_monitor.get("description", None)
         monitor_dict = dict()
         tenant, name = conv_utils.get_tenant_ref(name)
-
-        monitor_dict['tenant_ref'] = tenant_ref
+        if not tenant_ref == 'admin':
```
grastogi23: if tenant_ref != 'admin'

python/avi/f5_converter/persistence_converter.py
```diff
@@ -58,19 +58,21 @@ def convert(self, f5_config, avi_config, user_ignore, tenant_ref):
                 profile = prof_conv.update_with_default_profile(
                     persist_mode, profile, f5_persistence_dict, name)
                 tenant, name = conv_utils.get_tenant_ref(name)
+                if not tenant_ref == 'admin':
```
grastogi23: same here if tenant_ref != 'admin'

python/avi/f5_converter/persistence_converter.py
```diff
@@ -87,7 +89,7 @@ def convert(self, f5_config, avi_config, user_ignore, tenant_ref):
                         'profile', "hash-persistence", name, conv_status, msg)
                     continue
                 else:
-                    LOG.error(
+                    LOG.warn(
```
grastogi23: it should be LOG.warning not just warn

python/avi/netscaler_converter/setup.py
```diff
@@ -1,13 +1,14 @@
...
+AVI_PIP_VERSION = ''
+
 # allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
-AVI_PIP_VERSION = ''
 
 setup(
     name='avinetscalerconverter',
-    version=AVI_PIP_VERSION,
+    version= AVI_PIP_VERSION,
```
grastogi23: pl. check if space is important in the create_sdk_pkgs.

PR #224, opened 2017-03-23
Fixed: ssl cipher yaml file 

rameshtathe:
I have made the changes to fix the ssl cipher yaml file

### DIFF
--- a/python/avi/netscaler_converter/profile_converter.py
+++ b/python/avi/netscaler_converter/profile_converter.py
@@ -2,11 +2,10 @@ import logging
 import avi.netscaler_converter.ns_util as ns_util
 import os
 import re
-import yaml
 import avi.netscaler_converter.ns_constants as ns_constants
+
 from datetime import datetime
 from OpenSSL import crypto as c
-
 from avi.netscaler_converter.ns_constants import (STATUS_SKIPPED,
                                                   STATUS_SUCCESSFUL,
                                                   STATUS_INDIRECT,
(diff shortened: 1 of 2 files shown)


python/avi/netscaler_converter/profile_converter.py
```diff
@@ -72,8 +72,10 @@ def __init__(self, tenant_name, cloud_name, tenant_ref, cloud_ref,
         self.tenant_ref = tenant_ref
         self.cloud_ref = cloud_ref
         # ssl cipher yaml
-        ssl_ciphers_dict = yaml.safe_load(open(ssl_ciphers_yaml))
-        self.netscaler_ssl_cipher_to_open_ssl_cipher = ssl_ciphers_dict.get('netscaler_ssl_cipher_to_open_ssl_cipher', {})
+        ssl_ciphers_dict = yaml.safe_load(open(os.path.dirname(__file__) + '/%s'
```
grastogi23: This is not the right place for resolving the directory. ssl_ciphers_yaml was passed as argument. This file name should be resolved properly there i.e netscaler_config_converter.py:31

grastogi23 (changes requested):

python/avi/netscaler_converter/profile_converter.py
rameshtathe: Fixed.
load the ssl cipher yaml file in netscaler_config_converter.py and passed to profile converter

grastogi23 (approved):

PR #231, opened 2017-04-04
Changes includes:

rameshtathe:
1. Added appdirs dependancy in pip setup of sdk and migrationtools

### DIFF
--- a/python/avi/sdk/setup.py
+++ b/python/avi/sdk/setup.py
@@ -32,7 +32,7 @@ setup(
     keywords='AVI ADC Loadbalancer automation datacenter SDK',
     license='Avi Networks',
     include_package_data=True,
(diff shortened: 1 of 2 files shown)


python/avi/sdk/setup.py
```diff
@@ -32,7 +32,7 @@
     keywords='AVI ADC Loadbalancer automation datacenter SDK',
     license='Avi Networks',
     include_package_data=True,
-    install_requires=['requests'],
+    install_requires=['requests', 'appdirs'],
```
grastogi23: this should not be added. as only migration tools depends on appdirs.

grastogi23 (changes requested):

grastogi23 (approved):

PR #236, opened 2017-04-11
Netscaler & F5 Features

rameshtathe:
1. Convert default irules to policies instade of datascript
2. Netscaler Converter : SSL ciphers fails to upload to controller
3. Netscaler converter SSL and other profile merging
4. Fixed: The output count of successful conversions, but the pivot table does not show that same number
5. Fix issue with translate-port parameter in bigip config


python/avi/migrationtools/netscaler_converter/profile_converter.py
```diff
@@ -15,10 +15,19 @@
...
+    'app_profile': {},
+    'network_profile': {}
+}
+
+
 class ProfileConverter(object):
 
     def __init__(self, tenant_name, cloud_name, tenant_ref, cloud_ref,
-                 ssl_ciphers, keypassphrase=None):
+                 ssl_ciphers, profile_merge_check, keypassphrase=None):
```
grastogi23, line 30: argument description

grastogi23 (changes requested):

python/avi/migrationtools/netscaler_converter/profile_converter.py
rameshtathe: Done

grastogi23 (approved):

PR #237, opened 2017-04-11
Av 19361 referance tree

chaitanyaavi:


python/avi/migrationtools/vs_filter.py
```diff
@@ -29,50 +29,56 @@ def filter_for_vs(avi_config, vs_names):
...
             new_config[avi_conf_key].append(found_obj)
     else:
         new_config[avi_conf_key] = [found_obj]
-    find_and_add_objects(found_obj, avi_config, new_config)
+    depth += 1
+    find_and_add_objects(found_obj, avi_config, new_config, depth)
 
 
-def find_and_add_objects(object, avi_config, new_config):
+def find_and_add_objects(object, avi_config, new_config, depth):
```
grastogi23: object is not a good name to use for the field as there is a class named object. Pylint should flag this out.

python/avi/migrationtools/vs_filter.py
```diff
@@ -29,50 +29,56 @@ def filter_for_vs(avi_config, vs_names):
...
         vs = vs[0]
         new_config['VirtualService'].append(vs)
-        find_and_add_objects(vs, avi_config, new_config)
+        print '%s(VirtualService)' % vs_name
+        find_and_add_objects(vs, avi_config, new_config, depth=0)
     return new_config
 
 
-def search_obj(entity, name, new_config, avi_config):
+def search_obj(entity, name, new_config, avi_config, depth):
```
grastogi23, line 55: missing documentation

python/avi/migrationtools/vs_filter.py
```diff
@@ -95,14 +101,21 @@ def find_and_add_objects(object, avi_config, new_config):
...
-    print 'written avi config file ' + output_dir + \
-          os.path.sep + "FilterOutput.json"
+    if not args.print_only:
+        output_dir = os.path.normpath(args.output_file_path)
+        if not os.path.exists(args.output_file_path):
+            os.mkdir(args.output_file_path)
+        text_file = open(output_dir + os.path.sep + "FilterOutput.json", "w")
+        json.dump(new_avi_config, text_file, indent=4)
+        text_file.close()
+        print 'written avi config file ' + output_dir + \
```
grastogi23: with print you can use ',' rather than +. It will also make indentation easier.

grastogi23 (changes requested):

grastogi23 (approved):

PR #239, opened 2017-04-12
Update Ansible modules to support API Version

cjawale:


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -263,6 +263,8 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
             module.params['password'],
             tenant=module.params['tenant'])
     state = module.params['state']
+    # Get the api version.
+    api_version = module.params['api_version']
```
grastogi23: we should check if the module has api_Version because if the ansible module is old and sdk is new then this will result in exception.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -293,12 +297,17 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
     log.info('passed object %s ', obj)
 
     if name is not None:
-        existing_obj = api.get_object_by_name(
-            obj_type, name, tenant=tenant, tenant_uuid=tenant_uuid,
-            params={'include_refs': '', 'include_name': ''})
+        # added api version to avi api call.
+        existing_obj = api.get_object_by_name(obj_type, name, tenant=tenant,
+                                              tenant_uuid=tenant_uuid,
```
grastogi23: bad indentation...

grastogi23 (changes requested):

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -263,6 +263,10 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
             module.params['password'],
             tenant=module.params['tenant'])
     state = module.params['state']
+    # Get the api version.
+    api_version = module.params.get('api_version', None)
+    if not api_version:
+        api_version = '16.4'
```
grastogi23, line 269: can you pl. confirm that you have tested this case.

grastogi23 (approved):

PR #242, opened 2017-04-18
Changed version to 17.1

cjawale:


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -266,7 +266,7 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
             tenant=module.params['tenant'])
     state = module.params['state']
     # Get the api version.
-    api_version = module.params.get('api_version', '16.4')
+    api_version = module.params.get('api_version', '17.1')
```
grastogi23, line 269: no this should remain 16.4 not 17.1. only the check for version of sdk in ansible needs to change.

grastogi23 (changes requested):

cjawale:
Changes Already merged into master.

PR #252, opened 2017-05-03
Bug fix in avi_api.py

ypraveen:
Api version not properly passed down when connection gets reset.


grastogi23 (approved):
Need to also cherry-pick to 17.1.1

PR #254, opened 2017-05-04
Fixed referance issue and added support for tenants creation

chaitanyaavi:


python/avi/migrationtools/f5_converter/conversion_util.py
```diff
@@ -15,6 +15,7 @@
 
 LOG = logging.getLogger(__name__)
 csv_writer_dict_list = []
+tenants = []
```
grastogi23, line 18: should use set.

grastogi23: I will merge this for now.. but code will be much efficient that way

grastogi23 (approved):

PR #256, opened 2017-05-08
17.1.1 bug fixes

rameshtathe:
I have made the following changes:
1. Fixed issue of responder policy action for netscaler converter
2. Changed HM status SKIPPED to EXTERNAL those are not supported HM


python/avi/migrationtools/netscaler_converter/ns_util.py
```diff
@@ -1228,7 +1229,8 @@ def vs_per_skipped_setting_for_references(avi_config):
                         for http_req in \
                                 each_http_policy['http_request_policy'][
                                     'rules']:
-                            if http_req.get('switching_action'):
+                            if http_req.get('switching_action') and \
```
grastogi23, line 1232: these should have default as None too

grastogi23 (approved):

PR #258, opened 2017-05-09
Updating master for changes in 17.1.1

chaitanyaavi:


python/version.py
```diff
@@ -4,8 +4,8 @@
 @author: grastogi
 '''
 
-AVI_VERSION = '17.1'
-AVI_PIP_VERSION = AVI_VERSION 
+AVI_VERSION = '17.1.1'
```
grastogi23: this should not be merged...

grastogi23 (changes requested):

grastogi23 (approved):

PR #274, opened 2017-05-25
Added new changes for migration hardening

cjawale:


python/avi/migrationtools/avi_converter.py
```diff
@@ -58,10 +59,8 @@ def write_output(self, avi_config, output_dir, report_name, prefix):
         :param prefix: prefix for object
         :return: None
         """
-        if prefix:
```
grastogi23, line 61: seems like this getting rid of prefix.. is this expected?

grastogi23 (changes requested):

python/avi/migrationtools/avi_converter.py
cjawale: Yes, this should be remove, because these were not the pull changes which CD has created PR for 
there was slight confusion in branch name while creating PR.
So these are the conflict changes which was tested yesterday.

grastogi23 (approved):

PR #276, opened 2017-05-27
Added py test suite

cjawale:


grastogi23:
can we discuss if these packages need to be part of the migration tools...

kunalpatil501:
As per the discussion, removed the following packages from setup.py which were mainly needed for test automation :
1) pytest-html
2) pytest-cov
3) xlrd

Also updated the jenkins automation pipeline accordingly

PR #282, opened 2017-06-03
AV-21760: GSB: X-AVI-Version should be set to the right-value

nmandeyam:
=============================================================

There are 2 commits to this issue:

01. In avi-sdk layer, we will parse the login-data and extract the
    remote-sites api-version.

02. In the glb_mgr_sessions.py, we will do a min(my_version, remote_version)
    and use it in the X-Avi-Version.

Tests:
01. Ran gslb_site,txt, gslb_site_geo.txt, gslb_site_housekeeping.txt, gslb_site_fsm.txt
02. Ran upgrade multiple times to ensure that the system is consistent prior and after
    upgrade.

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -167,6 +167,7 @@ class ApiSession(Session):
         self.key = controller_ip + ":" + username
         self.api_version = api_version
         self.retry_conxn_errors = retry_conxn_errors
+        self.remote_api_version = None
 
         # Refer Notes 01 and 02
         if controller_ip.startswith('http'):
@@ -253,6 +254,18 @@ class ApiSession(Session):
         self.headers = {}
         self.authenticate_session()
 
+    def get_remote_api_version(self, rsp):
+        """
+        Retrieves the remote api version from the login-data.
+        """
+        login_data = rsp.json()
+        controller_info = login_data.get('version', None)
+        if controller_info:
+            version = controller_info.get('Version', None)
+            if version:
+                self.remote_api_version = str(version)
+        return
+
     def authenticate_session(self):
         """
         Performs session authentication with Avi controller and stores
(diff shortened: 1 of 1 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -253,6 +254,18 @@ def reset_session(self):
         self.headers = {}
         self.authenticate_session()
 
+    def get_remote_api_version(self, rsp):
```
grastogi23, line 257: let us add what rsp is. I think we should pass json version to this function. 
Let us also change the name to be cache_controller_api_version.

python/avi/sdk/avi_api.py
```diff
@@ -253,6 +254,18 @@ def reset_session(self):
...
+        """
+        Retrieves the remote api version from the login-data.
+        """
+        login_data = rsp.json()
+        controller_info = login_data.get('version', None)
+        if controller_info:
+            version = controller_info.get('Version', None)
+            if version:
+                self.remote_api_version = str(version)
```
grastogi23, line 266: no need to do str(version) as you can have a situation when default is None and that would be stored as "None" and would fail the parsing. You can also choose to initialize as empty string as "".

grastogi23: how about 
self.remote_api_versio = rsp.json().get('version', {})

This way in future we can use other information in the API version like min_version etc. It would mean in your code you will have to do 

session.remove_api_version.get('Version', '')

grastogi23 (changes requested):

python/avi/sdk/avi_api.py
```diff
@@ -268,9 +281,11 @@ def authenticate_session(self):
         rsp = super(ApiSession, self).post(self.prefix+"/login", body,
                                            timeout=self.timeout)
         if rsp.status_code != 200:
+            self.remote_api_version = None
```
grastogi23, line 284: change it to self.remote_api_version = {}

python/avi/sdk/avi_api.py
nmandeyam: Accept

python/avi/sdk/avi_api.py
nmandeyam: Accept

python/avi/sdk/avi_api.py
nmandeyam: Acccept

PR #285, opened 2017-06-05
Converter hardening 2

chaitanyaavi:


python/avi/migrationtools/ansible/library/avi_traffic.py
```diff
@@ -46,90 +47,101 @@ def main():
...
-                time.sleep(0.2)
-    if vs_up:
-    # Send http get requests to generate traffic
-        if request_type == 'http' or request_type == 'https':
+    if response.status_code >= 200 and response.status_code <= 299:
+        response = response.json()
+        uuid = response['results'][0]['uuid']
+        vs_up = False
+        if uuid:
+            status_path = '/api/virtualservice/%s/runtime/detail/' % uuid
```
grastogi23: optional one would be to use inventory... sometimes that is better than using runtime.

python/avi/migrationtools/ansible/library/avi_traffic.py
```diff
@@ -46,90 +47,101 @@ def main():
     response = requests.get(request_type + '://' + controller + path +
```
grastogi23: let us use the Avi ApiSession rather than requests library  for controller. This may become an issue later for openstack etc type environments.

grastogi23: Also, it expects basic auth which may not be the case always.

python/avi/migrationtools/ansible/library/avi_traffic.py
```diff
@@ -46,90 +47,101 @@ def main():
...
+        vs_up = False
+        if uuid:
+            status_path = '/api/virtualservice/%s/runtime/detail/' % uuid
             for i in range(0, 10):
+                # Check if virtualservice is up
+                response = requests.get(request_type + '://' + controller +
+                                        status_path, verify=False,
+                                        auth=(username, password))
+                response = response.json()
+                if 'OPER_UP' == response[0]['oper_status']['state']:
```
grastogi23: let us change to inventory. Also, tenant needs to be specified as well.

python/avi/migrationtools/ansible/library/avi_traffic.py
```diff
@@ -46,90 +47,101 @@ def main():
...
-            else:
-                module.exit_json(
-                    stderr='Dns Lookup unsuccessful',
-                    changed=False,
-                    success=1
-                )
     else:
         module.exit_json(
-            stderr='Virtualservice is not enabled',
+            stderr='Virtualservice api is not reachable',
```
grastogi23: I believe it should VIP not API

python/avi/migrationtools/f5_converter/conversion_util.py
```diff
@@ -515,7 +520,9 @@ def get_service_obj(destination, avi_config, enable_ssl, controller_version,
         port = 0
     if isinstance(port, str) and (not port.isdigit()):
         port = get_port_by_protocol(port)
-
+    # Port is None then skip vs
+    if not port:
+        return None, None, None
```
grastogi23: should have a log here with reason for skip. Might be needed in debugging.

python/avi/migrationtools/f5_converter/conversion_util.py
```diff
@@ -756,8 +763,12 @@ def get_snat_list_for_vs(snat_pool):
     elif isinstance(members, str):
         ips = [members]
     if None in ips:
-        ips.remove(None)
+        ips = filter(None, ips)
```
grastogi23: this does not seem right. If the goal is to remove any member that is not None then 
ips = [ip for ip in ips if ip]

python/avi/migrationtools/f5_converter/monitor_converter.py
```diff
@@ -158,8 +158,16 @@ def convert_monitor(self, f5_monitor, key, monitor_config, input_dir,
         elif monitor_type == "dns":
             na_list = self.na_dns
             u_ignore = user_ignore.get("dns", [])
-            skipped = self.convert_dns(monitor_dict, f5_monitor, skipped)
-            ignore_for_defaults.update({'qtype': 'a'})
+            if f5_monitor.get('qname', None) and (not f5_monitor.get(
+                    'qname') == 'none'):
```
grastogi23, line 162: 'none' does not seem right....

python/avi/migrationtools/f5_converter/pool_converter.py
```diff
@@ -400,8 +402,8 @@ def convert_servers_config(self, servers_config, nodes, avi_config):
                 'enabled': enabled,
                 'description': description,
             }
-            # Check if port is present
-            if port:
+            # Check if port is present and added check for port is digit.
+            if port and str(port).isdigit():
                 server_obj['port'] = port
```
grastogi23, line 407: add debug when condition is not true?

python/avi/migrationtools/f5_converter/vs_converter.py
```diff
@@ -302,26 +303,28 @@ def convert_vs(self, vs_name, f5_vs, vs_state, avi_config, snat_config,
...
-                    'Ignore the snat IPs, its count is beyond 32 for vs : %s' %
-                    vs_name)
-            else:
-                vs_obj["snat_ip"] = snat_list
-            conv_status = {'status': final.STATUS_SUCCESSFUL}
-            message = 'Mapped indirectly to VS -> SNAT IP Address'
-            conv_utils.add_conv_status('snatpool', '', snat_pool_name,
-                                       conv_status, message)
+        # Checking snat conversion flag for creating snat ip object
+        if self.con_snatpool:
```
grastogi23: we should may be print message if  snat = f5_vs.get("source-address-translation", {}) is not empty and flag self.con_snatpool is set. This way information would be there in log as why snat was skipped.

grastogi23 (changes requested):

python/avi/migrationtools/f5_converter/pool_converter.py
cjawale: Done. Added debug log when condition is not true.

python/avi/migrationtools/f5_converter/monitor_converter.py
chaitanyaavi: F5 input file may have string value as 'none' added this condition to avoid that scenario.

grastogi23 (approved):

PR #289, opened 2017-06-08
AV-21986 Fixed Migration tool creates additional unused pool objects …

chaitanyaavi:
…during netscaler conversion and added support for back pool with poolgroup


grastogi23 (approved):

grastogi23:
AV-21986 #commit

PR #290, opened 2017-06-12
Sample to clone virtual service, pool groups or pools

thebertster:
Proposed sample script to clone VS, pool groups or pools, optionally to a different tenant and/or cloud (where possible).

Existing sample script for virtual services includes a clone function but that does not handle anything except a basic use case.


python/avi/sdk/samples/AviClone.py
```diff
@@ -0,0 +1,734 @@
+#!/usr/bin/env python
```
grastogi23: we are following Google Guidelines  + PeP8 for the coding where the file names are small letters with words separated by "_". It would be great if you can do another pull request with the file name change.

python/avi/sdk/samples/AviClone.py
```diff
@@ -0,0 +1,734 @@
...
+
+class AviClone:
+    def __init__(self, api):
+        self.api = api
+        self.flush_actions()
+
+    def flush_actions(self):
+        self.actions = []
+
+    def get_all_objects_by_name(self, path, name, tenant='', tenant_uuid='',
```
grastogi23: It would be good to later contribute to the main ApiSession as a get all method.

python/avi/sdk/samples/AviClone.py
```diff
@@ -0,0 +1,734 @@
...
+                        params=params, api_version=api_version, **kwargs)
+        if resp.status_code in (401, 419):
+            ApiSession.reset_session(self.api)
+            resp = self.api.get_object_by_name(
+                    path, name, tenant, tenant_uuid, timeout=timeout,
+                    params=params, **kwargs)
+        if resp.status_code < 300:
+            obj = resp.json()['results']
+
+        self.api._update_session_last_used()
```
grastogi23: no need to do this as it is automatically done inside self.api.get

python/avi/sdk/samples/AviClone.py
```diff
@@ -0,0 +1,734 @@
...
+
+        obj = None
+        if not params:
+            params = {}
+        params['name'] = name
+        resp = self.api.get(path, tenant, tenant_uuid, timeout=timeout,
+                        params=params, api_version=api_version, **kwargs)
+        if resp.status_code in (401, 419):
+            ApiSession.reset_session(self.api)
+            resp = self.api.get_object_by_name(
```
grastogi23: you can simply do resp.json(). That will automatically raise exception if it is 401 or 419 case.

thebertster:
Do we think it is worth adding the get_all_objects_by_name() method to the main SDK? The existing get_object_by_name() only returns the first match even though it's possible for there to be objects with the same name in different clouds (e.g. for SE groups).

grastogi23 (approved):

PR #309, opened 2017-06-30
AV23102 Fixed the ssl key cert ref issue for pool

Rishabh-Baranwal:


grastogi23:
I don't see unit test case with this?

PR #323, opened 2017-07-11
Av 23449 : converted avi config for ssl profiles incorrect - Fixed

chaitanyaavi:
Changed ssl_profile_name to ssl_profile_ref


python/avi/migrationtools/netscaler_converter/ns_util.py
```diff
@@ -1087,23 +1087,6 @@ def get_network_profile_skipped(csv_writer_dict_list, name_of_object, vs_ref):
     return network_profile_name, skipped_list
```
chaitanyaavi, line 1089: Removed as duplicate function might have got duplicated while merging.

grastogi23 (approved):

PR #327, opened 2017-07-18
Av 23610 : Added progress bar for F5 and netscaler

chaitanyaavi:


python/avi/migrationtools/avi_converter.py
```diff
@@ -8,6 +8,31 @@
 
 LOG = logging.getLogger(__name__)
 
+
+# Print iterations progress
+def print_progress_bar(iteration, total, msg, prefix='', suffix='', decimals=1,
+                       length=100, fill='#'):
```
grastogi23: should length be 80 for 80 char window. or may be just 50 so it does not wrap around?

python/avi/migrationtools/config_patch.py
```diff
@@ -201,12 +202,20 @@ def patch(self):
         :return:
         """
         new_cfg = deepcopy(self.avi_cfg)
+        total_size = len(self.patches)
+        progressbar_count = 0
+        print "Conversion For Patching of Object started..."
```
grastogi23: Patching of objects

grastogi23 (changes requested):

python/avi/migrationtools/config_patch.py
```diff
@@ -201,12 +202,20 @@ def patch(self):
         :return:
         """
         new_cfg = deepcopy(self.avi_cfg)
+        total_size = len(self.patches)
+        progressbar_count = 0
+        print "Conversion For Patching of objects started..."
+        from avi.migrationtools.avi_converter import print_progress_bar
```
grastogi23: why we are importing it here?

python/avi/migrationtools/gss_convertor/gss_parser.py
```diff
@@ -30,7 +30,7 @@ def parse(val):
 
 def parser(file_name):
     """ Parsing goes here """
-    LOG.info('Parser Started')
+    LOG.info('Parserend Started')
```
grastogi23: spelling

grastogi23 (changes requested):

python/avi/migrationtools/config_patch.py
chaitanyaavi: There is cyclic dependency on config_patch.py and avi_converter.py hence we had to import it here.

grastogi23: let us then move the print_progress_bar into a common migration_utils.py. we can use this to put any common utility that is needed across different modules.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -0,0 +1,3324 @@
...
+    # 'ssl key and cert': {},
+    # 'ssl profile': {},
+    # 'pool group': {},
+    # 'health monitor': {},
+    # 'Httppolicy': {}
+}
+
+
+
+class Migration_Utils(object):
```
grastogi23: we don't use _ in the class names. so it should be MigrationUtils

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -0,0 +1,3324 @@
...
+            format(100 * (iteration / float(total)))
+        filledLength = int(length * iteration // total)
+        bar = fill * filledLength + '-' * (length - filledLength)
+        if (iteration < total):
+            print '\r%s |%s| %s%% %s' % (prefix, bar, percent, suffix),
+        else:
+            print '\r%s |%s| %s%% %s' % (prefix, bar, percent, suffix)
+
+
+class Ns_Migration_Utils(Migration_Utils):
```
grastogi23: Ns_Migration_Utils -> NsMigrationUtils

grastogi23 (changes requested):

grastogi23 (approved):

PR #328, opened 2017-07-18
AV23642: Baseprofile merge review comment.

Rishabh-Baranwal:


python/avi/migrationtools/f5_converter/f5_config_converter.py
```diff
@@ -7,15 +7,25 @@
...
 import os
 import json
 
 
 LOG = logging.getLogger(__name__)
 csv_writer = None
-
+# Define Dict of merge_object_mapping to update the merged monitor, profile
+# name of ssl_profile, application_profile, network_profile etc
+merge_object_mapping = {
```
grastogi23, line 21: can this be generic... there are other profiles too that can benefit from merging.

python/avi/migrationtools/f5_converter/f5_config_converter.py
```diff
@@ -41,54 +51,68 @@ def convert(f5_config, output_dir, vs_state, input_dir, version,
     """
 
     avi_config_dict = {}
+    sys_dict = {}
     try:
         # load the yaml file attribute in f5_attributes.
         f5_attributes = conv_const.init(version)
+        sys_dict['ApplicationProfile'] = []
```
grastogi23: also do 
merge_obj_types = ['NetworkProfile', 'SSLProfile'...]
for obj_type in merge_obj_types.keys():
   sys_dict[obj_type] = []

python/avi/migrationtools/f5_converter/f5_config_converter.py
```diff
@@ -41,54 +51,68 @@ def convert(f5_config, output_dir, vs_state, input_dir, version,
...
+        sys_dict['NetworkProfile'] = []
+        sys_dict["SSLProfile"] = []
+        sys_dict['PKIProfile'] = []
+        sys_dict['ApplicationPersistenceProfile'] = []
+        sys_dict['HealthMonitor'] = []
         if profile_path and os.path.exists(profile_path):
             with open(profile_path) as data:
                 prof_data = json.load(data)
-                avi_config_dict['ApplicationProfile'] = \
+                sys_dict['ApplicationProfile'] = \
```
grastogi23: again this can be 
for k in merge_obj_types.keys()
   sys_dict[k] = prof_data.get(k, [])

python/avi/migrationtools/f5_converter/f5_config_converter.py
```diff
@@ -41,54 +51,68 @@ def convert(f5_config, output_dir, vs_state, input_dir, version,
...
-            avi_config_dict['ApplicationProfile'] = []
-            avi_config_dict['NetworkProfile'] = []
-            avi_config_dict["SSLProfile"] = []
-            avi_config_dict['PKIProfile'] = []
-            avi_config_dict['ApplicationPersistenceProfile'] = []
-            avi_config_dict['HealthMonitor'] = []
+
+        avi_config_dict['ApplicationProfile'] = []
+        avi_config_dict['NetworkProfile'] = []
+        avi_config_dict["SSLProfile"] = []
```
grastogi23: this can be simplified too.

python/avi/migrationtools/netscaler_converter/netscaler_config_converter.py
```diff
@@ -61,60 +62,68 @@ def convert(meta, ns_config_dict, tenant_name, cloud_name, version, output_dir,
         avi_config['META']['supported_migrations']['versions'].append(
             'current_version')
 
+        sys_dict['ApplicationProfile'] = []
+        sys_dict['NetworkProfile'] = []
+        sys_dict["SSLProfile"] = []
```
grastogi23: this code should be common in some place...

python/avi/migrationtools/netscaler_converter/ns_util.py
```diff
@@ -679,26 +680,33 @@ def create_self_signed_cert():
...
+        ob_cp = copy.deepcopy(obj)
+        del ob_cp["name"]
+        if "description" in ob_cp:
+            del ob_cp["description"]
+        if 'url' in ob_cp:
+            del ob_cp['url']
+        if 'uuid' in ob_cp:
+            del ob_cp['uuid']
+        if cmp(src_cp, ob_cp) == 0:
+            return obj["name"], src_obj['name']
```
grastogi23: name matching may not be sufficient as two netscalers conf may have different objects with same name.

python/avi/migrationtools/netscaler_converter/ns_util.py
```diff
@@ -679,26 +680,33 @@ def create_self_signed_cert():
...
     for tmp_obj in obj_list:
-        src_cp = copy.deepcopy(src_obj)
         tmp_cp = copy.deepcopy(tmp_obj)
-        del src_cp["name"]
-        if "description" in src_cp:
-            del src_cp["description"]
         del tmp_cp["name"]
```
grastogi23: you can use tmp_cp.pop instead. It has option to ignore check for keys.

grastogi23 (changes requested):
Hi
Let us review the object comparison. as it seems like it is only happening using name which is not sufficient.

python/avi/migrationtools/netscaler_converter/ns_util.py
Rishabh-Baranwal: Sorry Gaurav, correct me if am wrong. We merge profile/monitor on basis of all attributes/keys except 'description' and 'name' as we delete it before comparison.

Rishabh-Baranwal: Also for System profiles we remove 'url' and 'uuid' keys/attributes before comparison

python/avi/migrationtools/f5_converter/f5_config_converter.py
Rishabh-Baranwal: the dict is having all entities which comes as a result from conversion and needs merging. I agree with you but can we go ahead with this as of now , will check the possibility.

grastogi23 (approved):

grastogi23:
AV-23642 #commit

PR #330, opened 2017-07-20
Added fix for preserving the user custom headers. Also, fixed bug whe…

grastogi23:
…re content-type header was not preserved after session reset


ypraveen (approved):

grastogi23:
AV-23988 #commit

PR #348, opened 2017-08-04
AV-22308:Redirect HTTP VS to HTTPS VS if there is no pool, and remove HTTP VS.

Rishabh-Baranwal:


grastogi23:
@Rishabh-Baranwal pl. update to 17.1.5

grastogi23 (approved):

PR #353, opened 2017-08-09
AV 21016: Validation method completed.

Rishabh-Baranwal:


grastogi23:
@Rishabh-Baranwal pl. rebase to 17.1.6

grastogi23:
@Rishabh-Baranwal  pl. add unit tests.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -210,4 +211,171 @@ def print_progress_bar(self, iteration, total, msg, prefix='', suffix='',
...
+        """
+        :param entity_names: list of name of the avi entity/object
+        :param prop_name: property name
+        :param value: property value
+        :param limit_data: data from generated yaml
+        :param obj: obj name
+        :return: validity and new value
+        """
+        valid = None
+        new_value = value if value is not None else None
```
grastogi23: why check for None. you can simply do new_value = value. It would be the same behavior.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -210,4 +211,171 @@ def print_progress_bar(self, iteration, total, msg, prefix='', suffix='',
...
+        for key, val in limit_data.iteritems():
+            pr = val.get(obj, {})
+            if not pr:
+                continue
+            LOG.debug("Validating property '%s'" % ('-->'.join(entity_names) +
```
grastogi23: use the format style debugging. This will result in evaluation of arguments regardless of whether debug is turned on or off.
eg. log.debug("%s", x) instead of log.debug("%s" % x)

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -227,4 +228,171 @@ def print_progress_bar(self, iteration, total, msg, prefix='', suffix='',
...
+            if typ:
+                break
+        else:
+            LOG.debug("Property '%s' is not present in generated json, reason "
+                   "being the property doesn't have any attribute from the "
+                   "list %s" % (prop_name, str(['default_value', 'range',
+                                                'special_values', 'ref_type'])))
+            return None, None
+        if new_value is not None:
+            if type(new_value) == unicode:
```
grastogi23: may want to check for basestring as well.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -227,4 +228,171 @@ def print_progress_bar(self, iteration, total, msg, prefix='', suffix='',
...
+                        valid = False
+                        new_value = p_key.get('default_value')
+                    else:
+                        valid = True
+                        LOG.debug("Value '%s' is fine" % str(new_value))
+                if typ == 'bool':
+                    if new_value not in (False, True, 'False', 'True'):
+                        valid = False
+                        new_value = p_key.get('default_value')
+                        new_value = False if new_value == 'False' else True
```
grastogi23, line 291: what if new_value == False instead of 'False'. this logic would make it True.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -227,4 +228,171 @@ def print_progress_bar(self, iteration, total, msg, prefix='', suffix='',
...
+                    else:
+                        valid = True
+                        LOG.debug("Value '%s' is fine" % str(new_value))
+            else:
+                LOG.debug("Type of value '%s' doesn't match with type '%s' "
+                       "defined" % (str(type(new_value)), typ))
+                valid, new_value = None, None
+        else:
+            if eval(p_key.get('required')):
```
grastogi23: why eval?

grastogi23 (changes requested):

python/avi/migrationtools/avi_migration_utils.py
Rishabh-Baranwal: sorry, I didn't realize that I have a cover made by me at line 260, will remove that.

python/avi/migrationtools/avi_migration_utils.py
Rishabh-Baranwal: Ok, will do that for every debug line.

python/avi/migrationtools/avi_migration_utils.py
Rishabh-Baranwal: I hadn't seen python type returning from protobuf as unicode or basestring, had seen 'str' only by far. Hence, need to encode the unicode value to make it python type 'str'. If I use basestring check, it will be true for both unicode and str then I will not be able to decide which one to encode to make it proper str.

python/avi/migrationtools/avi_migration_utils.py
Rishabh-Baranwal: I believe, new_value can't be False as p_key.get('default_value') will give either 'True' or 'False'.

python/avi/migrationtools/avi_migration_utils.py
Rishabh-Baranwal: ok, I will check for 'True' directly.

PR #359, opened 2017-08-17
Bug fixes for 1)Accepted cipher in ssl profile, 2)AV-25188, 3)Default SSL profile for https monitor, 4)Converting L4 VS to HTTP if service has ssl enabled.

Rishabh-Baranwal:


grastogi23:
AV-25188 #commit

PR #362, opened 2017-08-18
Av 24184 : Avi SDK should not print output in case the result contains a file ( binary )

chaitanyaavi:
…ns a file ( binary )


python/avi/sdk/avi_api.py
```diff
@@ -369,9 +369,16 @@ def _api(self, api_name, path, tenant, tenant_uuid, data=None,
             logger.error('Error in Requests library %s', e)
             raise
         if not connection_error:
-            logger.debug(
-                'path: %s http_method: %s hdrs: %s params: %s data: %s rsp: %s',
-                fullpath, api_name.upper(), api_hdrs, kwargs, data, resp.text)
+            try:
+                responses = resp.json()
```
grastogi23: I believe you can check if encoding of the response.

grastogi23 (changes requested):

python/avi/sdk/avi_api.py
chaitanyaavi: encoding returned in response is None in both the cases. 
We can change the logging to log the count of objects instead of whole response this also logs the response of full system export which has sensitive information like passwords etc.

PR #367, opened 2017-08-18
Added some additional functions

thebertster:
Added functions to allow cloning of generic objects of any type (provided they have no child references that need to be cloned).

Also added ability when cloning VS or pool between tenants to handle child objects which may be defined at the tenant level, such as application profile, persistence profile etc.


grastogi23 (approved):

grastogi23:
@thebertster this has conflicts. pl. resolve

PR #379, opened 2017-08-28
AV-25616 #commit: Fixed F5 HTTP vips being incorrectly given fastL4 app profi…

chaitanyaavi:
…le during conversion


grastogi23 (commented):
AV-25616 #commit

PR #380, opened 2017-08-28
AV-25105 #commit: Only client-ip persistence is applicable for Layer-4 virtual service but some VS having SSL application persistence profile. Also skipping VS when we have OC profile with no ssl profile but ssl persistence profile.

Rishabh-Baranwal:


grastogi23 (approved):

grastogi23:
AV-25105 #commit

PR #381, opened 2017-08-29
AV-25272: Added 3 characters random string in Merged name inorder to not have any conflict in names when we use a converted json as baseline.

Rishabh-Baranwal:


python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -126,10 +127,10 @@ def check_for_duplicates(self, src_obj, obj_list, obj_type, merge_object_mapping
                 if tmp_obj["name"] in merge_object_mapping[obj_type].keys():
                     merge_object_mapping[obj_type]['no'] += 1
                     no = merge_object_mapping[obj_type]['no']
-                    mid_name = ent_type and (
-                    'Merged-' + ent_type + '-' + obj_type
-                    + '-' + str(no)) or ('Merged-' +
-                                         obj_type + '-' + str(no))
+                    mid_name = ent_type and ('Merged-' + ent_type + '-' +
```
grastogi23: better to use format string rather than +. it is very hard to understand what is going on.

grastogi23 (changes requested):

grastogi23 (approved):

PR #383, opened 2017-08-30
AV-25748 fixed Converted Netscaler persistence timer needs to be set …

chaitanyaavi:
…to two mins if no timer is specified


grastogi23:
AV-25748 #commit

grastogi23 (approved):

PR #394, opened 2017-09-01
Validationlog:Added the hierarchy for object attributes to validation log

Rishabh-Baranwal:


python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -241,12 +242,15 @@ def validate_value(self, entity_names, prop_name, value, limit_data, obj):
         """
         valid = None
         new_value = value
+        msgvar = valname and entity_names and '%s-->%s-->%s' % (valname,
```
grastogi23, line 245: It is ok here but use of msgvar is only for debug printing then it should be with the debug print only. This way the args are not evaluated unless the debugs are turned on.

python/avi/migrationtools/avi_migration_utils.py
```diff
@@ -365,33 +370,47 @@ def validate_prop(self, dictval, heir, limit_data, obj):
                      'pki_profile_ref', 'pool_ref', 'pool_group_ref',
                      'http_policy_set_ref', 'ssl_key_and_certificate_refs',
                      'vsvip_ref', 'description']:
-                LOG.debug("Skipping validation checks for '%s'", k)
+                msvar = valname and heir and '%s-->%s-->%s' % (valname,
```
grastogi23, line 373: same here.

grastogi23 (approved):

PR #395, opened 2017-09-01
Ansible meta order Change

Jeyanthinath:


grastogi23:
We shouldn't check in the automatically generated files. let us discuss how to handle its integration.

python/avi/migrationtools/ansible/ansible_config_converter.py
```diff
@@ -43,6 +42,8 @@ class AviAnsibleConverter(object):
                                           'ansible_order_constant.yml')
     with open(ansible_rest_file_path, 'r') as f:
         default_meta_order = yaml.load(f)
+    default_meta_order = default_meta_order['avi_resource_types']
```
grastogi23: why are we using class variables here. They should be instance variables.  Pl. move to the __init__ function.

grastogi23 (changes requested):

grastogi23 (approved):

PR #397, opened 2017-09-05
Av 25877 Unit test case to automate clean reboot controller and upload license file.

patilshrikant486:
This PR contains:
1. Test case to automate clean reboot controller and upload license file.Now we can clean reboot controller every time before auto upload test cases for both F5 and Netscaler.
2. Test cases for Ansible object auto upload on the reboot cleaned controller.


python/avi/migrationtools/test/common/test_clean_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+    print res
+
+
+def upload_license(session):
+    """""
+    Upload the AviInternal license using API API And Verify
+    that it is uploaded successfully using assert.
+    """
+    response = session.put('license', data=json.dumps({
+        "license_text": "-----BEGIN PGP MESSAGE-----\r\nVersion: GnuPG/MacGPG2 v2.0.22 (Darwin)\r\nComment: GPGTools - https://gpgtools.org\r\n\r\nowGbwMvMwMWY/dmudiGn+0HGtYxlSQI5mcmpecWphgYG2XqVibk5IQ3L3kPF4jNT\r\nrBQcyzI980pSi/ISc7jKUo"
```
grastogi23: This is not good. We need to cleanup. we can't have our internal license on the web.

grastogi23 (changes requested):

PR #404, opened 2017-09-12
making default version correct rather than 16.4 which is considered i…

grastogi23:
…nvalid


imayobrown:
Thanks for the change

imayobrown (approved):

khaltore (approved):

PR #435, opened 2017-09-29
fix: AV-24465 , Excel sheet validation for merged objects 

Jeyanthinath:
Excel sheet validation with output json and merged objects with log file

### DIFF
--- a/python/avi/migrationtools/test/common/excel_reader.py
+++ b/python/avi/migrationtools/test/common/excel_reader.py
@@ -54,34 +54,31 @@ def output_sanitization(path_to_excel, path_to_out_json=None, path_to_log=None):
             if entity <> 'META' and entity <> 'VsVip':
                 for obj in file_strem[entity]:
                     out_obj.append(obj.get('name'))
-    # print len(out_obj)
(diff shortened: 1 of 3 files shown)


python/avi/migrationtools/test/common/excel_reader.py
```diff
@@ -38,22 +52,45 @@ def output_sanitization(path_to_excel, path_to_out_json=None):
...
+            a = file_strem.readlines()
+            b = str(a).split('$$$$$$')[-2].replace('\'', '"')
+            c = eval(b)
+
+    count = 0
+    obj_list = list()
     for obj in excel_obj:
         if obj not in out_obj:
-            print obj
+            flag = 1
```
grastogi23: Logic is not very clear. If the goal is to check the set difference then you can do comparison of set(obj) and set(c.keys().keys())....

grastogi23 (changes requested):

python/avi/migrationtools/test/common/excel_reader.py
Jeyanthinath: fixed

PR #437, opened 2017-09-29
added flag for enable_vs during converions, changed default to false

Jeyanthinath:


python/avi/migrationtools/ace_converter/ace_config_converter.py
```diff
@@ -17,13 +17,14 @@
 class ConfigConverter(object):
     """ Configuration conversion happens here """
 
-    def __init__(self, parsed_output, in_file, version='17.1.1'):
+    def __init__(self, parsed_output, in_file, enable_vs='False', version='17.1.1'):
```
grastogi23: enable_vs = False not within quotes. as if 'False' will be true.

grastogi23 (changes requested):

grastogi23 (approved):

PR #440, opened 2017-10-04
Bugfixes 17.2.3: Defaulted to LEASTCONNECTIONS for LB Algo, changed the extension from yml to yaml and added example for patch in F5 and NS. Also generated avi object key map based on meta order .

Rishabh-Baranwal:


grastogi23 (changes requested):
The unit tests are still referring to patch.yml. they need  to be changed to look into patch.yaml

Rishabh-Baranwal:
Correction done and travis also passed, hence removed the 'do not merge' label.

grastogi23 (approved):

PR #442, opened 2017-10-05
AV-27828 Added option to use uuid when it is present in object to per…

chaitanyaavi:
…form GET: Ansible

### DIFF
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -339,11 +339,8 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
                                    params={'include_refs': '', 'include_name': ''},
                                    api_version=api_version)
             existing_obj = existing_obj.json()
-        except:
-            if existing_obj.status_code == 404:
-                existing_obj = None
-            else:
-                return module.fail_json(msg=existing_obj.text)
+        except ObjectNotFound:
+            existing_obj = None
     elif name:
         params = {'include_refs': '', 'include_name': ''}
         if obj.get('cloud_ref', None):
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -340,8 +340,10 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
                                    api_version=api_version)
             existing_obj = existing_obj.json()
         except:
-            log.error('Failed to get avi object for uuid : %s' % (uuid))
-            return module.fail_json(msg=existing_obj.text)
+            if existing_obj.status_code == 404:
+                existing_obj = None
+            else:
+                return module.fail_json(msg=existing_obj.text)
     elif name:
         params = {'include_refs': '', 'include_name': ''}
         if obj.get('cloud_ref', None):
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -337,10 +337,11 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
         try:
             existing_obj = api.get(obj_path, tenant=tenant, tenant_uuid=tenant_uuid,
                                    params={'include_refs': '', 'include_name': ''},
-                                   api_version=api_version).json()
+                                   api_version=api_version)
+            existing_obj = existing_obj.json()
         except:
             log.error('Failed to get avi object for uuid : %s' % (uuid))
-            return
+            return module.fail_json(msg=existing_obj.text)
     elif name:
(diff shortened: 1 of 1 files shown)


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -295,9 +295,13 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
     api_version = api_creds.api_version
     name = module.params.get('name', None)
+    # Added Support to get uuid
+    uuid = module.params.get('uuid', None)
     check_mode = module.check_mode
     obj_path = None
-    if name is None:
+    if uuid:
+        obj_path = '%s/%s' % (obj_type, uuid)
+    elif not name and not uuid:
```
grastogi23: no need to test for not uuid as it will never come here if uuid was set.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -328,7 +332,12 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
 
     log.info('passed object %s ', obj)
 
-    if name is not None:
+    if uuid:
+        # Get the object based on uuid.
+        existing_obj = api.get(obj_path, tenant=tenant, tenant_uuid=tenant_uuid,
```
grastogi23: we may need to test for exception here. because if object is not present the .json() may result in failure.

grastogi23 (changes requested):

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -328,7 +332,16 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
-    if name is not None:
+    if uuid:
+        # Get the object based on uuid.
+        try:
+            existing_obj = api.get(obj_path, tenant=tenant, tenant_uuid=tenant_uuid,
+                                   params={'include_refs': '', 'include_name': ''},
+                                   api_version=api_version).json()
+        except:
+            log.error('Failed to get avi object for uuid : %s' % (uuid))
+            return
```
grastogi23: you can't return from here. It has to be 
return module.fail_json(msg=rsp.text)

try to create this scenario using playbook as test to make sure we get intended behavior.

grastogi23 (changes requested):

python/avi/sdk/utils/ansible_utils.py
chaitanyaavi: Tested with return module.fail_json(msg=rsp.text) for incorrect uuid. Got following response on console.

TASK [Create or Update virtualservice: triageavailability-prdap-lv.cfp.intuit.net_8080] *******************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "{\"error\": \"VirtualService object not found!\"}"}
	to retry, use: --limit @/home/vinay/avi_config_create_object.retry

PLAY RECAP ************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1  

Is it expected behavior?

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -328,7 +332,17 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
+    if uuid:
+        # Get the object based on uuid.
+        try:
+            existing_obj = api.get(obj_path, tenant=tenant, tenant_uuid=tenant_uuid,
+                                   params={'include_refs': '', 'include_name': ''},
+                                   api_version=api_version)
+            existing_obj = existing_obj.json()
+        except:
+            log.error('Failed to get avi object for uuid : %s' % (uuid))
+            return module.fail_json(msg=existing_obj.text)
```
grastogi23: I reviewed the code again. I believe here it should simply be except for object not found should be existing_obj = None. this way later part of the code can do post as object was not found.

grastogi23 (changes requested):

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -328,7 +332,19 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
     log.info('passed object %s ', obj)
 
-    if name is not None:
+    if uuid:
+        # Get the object based on uuid.
+        try:
+            existing_obj = api.get(obj_path, tenant=tenant, tenant_uuid=tenant_uuid,
+                                   params={'include_refs': '', 'include_name': ''},
+                                   api_version=api_version)
+            existing_obj = existing_obj.json()
```
grastogi23, line 341: It is better to catch the special object not found exception defined in the avi_api.py.

grastogi23 (changes requested):

grastogi23 (approved):

PR #450, opened 2017-10-06
fixes for duplicate domains in Gss

Jeyanthinath:


python/avi/migrationtools/gss_convertor/gss_config_convertor.py
```diff
@@ -207,6 +197,71 @@ def config_converter(file_name):
     set_total_stats('hang', hang_count)
     set_total_stats('no_member', no_member_count)
 
+    # fix for multiple az
```
grastogi23, line 200: can we break this into a different function. Too long makes it hard to read.

grastogi23 (changes requested):

python/avi/migrationtools/gss_convertor/gss_config_convertor.py
Jeyanthinath: ok

Jeyanthinath: fixed

grastogi23 (approved):

PR #454, opened 2017-10-09
AV-28004:Removing timeout if value=0

Rishabh-Baranwal:


grastogi23:
Let us keep the bug open and we need to fix the validation.

grastogi23 (approved):

PR #462, opened 2017-10-10
AV-23893: SDK Changes

chaitanyaavi:

PR #466, opened 2017-10-10
Av 27817 Rename ApiSession attribute keystone_token to token

nparikh-avinetworks:


python/avi/sdk/avi_api.py
```diff
@@ -202,7 +202,7 @@ def __init__(self, controller_ip, username, password=None, token=None,
         self.controller_ip = controller_ip
         self.username = username
         self.password = password
-        self.keystone_token = token
+        self.token = token
```
grastogi23, line 205: we shouldn't change the name... I am not sure if there is code that is accessing it with that name. 
@ypraveen: Added him for review on this as his os and heat plugins might be using it.

python/avi/sdk/avi_api.py
```diff
@@ -311,12 +311,11 @@ def authenticate_session(self):
...
-        body = {"username": self.username}
-        if not self.keystone_token:
-            body["password"] = self.password
-        else:
-            body["token"] = self.keystone_token
-
+        body = {
+            "username": self.username,
+            "password": self.password,
+            "token": self.token,
```
grastogi23: Was this causing problems? Isn't it better to not send empty strings for password or token?

python/avi/sdk/samples/cust_scripts/server_down_script.py
```diff
@@ -76,31 +76,31 @@ def Login(ws_addr='https://localhost', tenant='admin', user='admin',
...
-                            timeout=timeout, cookies=cookies,
-                            headers=headers, data=json.dumps(login_req))
+    # Either a password or a token is requiered (not both)
+    login_req = {
+        'username' : user,
+        'password' : password,
+        'token'    : token
+    }
+
+    rsp = requests.post(os.path.join(ws_addr, 'login'), verify=False,
```
grastogi23: let us convert this to use the SDK itself. Anyway this bad code and should actually be deleted.

grastogi23 (changes requested):

grastogi23:
@chaitanyaavi Added CD as he is bringing sweeping changes to SDK and that would need to be merged with this change.

python/avi/sdk/avi_api.py
ypraveen: I am not accessing keystone_token directly.

python/avi/sdk/avi_api.py
nparikh-avinetworks: It wasn't causing problems as written, but all such conditional logic for auth credentials has been removed from avi-dev. We modified the permissions code to put the onus of deciding which credentials to use on the authentication logic allowing empty strings and None values to be  handled gracefully by default.

python/avi/sdk/avi_api.py
nparikh-avinetworks: @grastogi23 @ypraveen 
Given that `keystone_token` doesn't appear to be used directly is it ok to keep this change? I renamed this attr because `token`  is being used for both keystone and API token auth. (I realize though that we are not the only consumers of this SDK so I should be more careful about this kind of change in the future).

python/avi/sdk/samples/cust_scripts/server_down_script.py
nparikh-avinetworks: I see what you mean. I assume the preferred way to login is by creating a valid ApiSession. Should I modify the example to use an ApiSession, delete the example, or make a different change here?

python/avi/sdk/avi_api.py
ypraveen: I am fine with renaming it. "token" is better than "keystone_token".

grastogi23: This field is anyway getting deprecated. https://github.com/avinetworks/sdk/pull/462. I would leave this change. I will ask CD to provide property for token. I am assuming there is not much folks in OS using API SDK directly.

python/avi/sdk/avi_api.py
grastogi23: but SDK could be used cross versions. I would pass on this change as we have many situations where customers use newer SDK with different versions of controller.

python/avi/sdk/samples/cust_scripts/server_down_script.py
grastogi23: If you are enhancing the code then change everything to ApiSession as requests library do not handle things like CSRF token etc. Otherwise, I will leave it up to you based on your time to modify it.

PR #478, opened 2017-10-25
Control Script to support Host-Header based switching (for GE DC/OS)

adityats:


python/avi/sdk/samples/cust_scripts/ge-dcos-script.py
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/python
+import copy, json, os, string, sys
+from avi.sdk.avi_api import ApiSession
+from avi.rest.url_utils import slug_from_uri
```
grastogi23: avi.rest.url_utils this is not packaged with the SDK. It will cause failure in pip package creation. Suggest commit this into the devops repository.

grastogi23 (changes requested):

adityats:
@grastogi23 Made the library call local to the script for now.

grastogi23 (approved):

PR #481, opened 2017-10-26
Av 20903 - Adding Support for traffic generation in F5 Converter with test_vip

Jeyanthinath:
new flag test_vip is added 
        * tested for f5
        * not tested in netscalar


python/avi/migrationtools/f5_converter/f5_converter.py
```diff
@@ -151,28 +154,29 @@ def convert(self):
...
                 not_supported_list_partition = not_supported_list_partition \
-                                               + not_supported_list
+                    + not_supported_list
                 self.dict_merge(partition_conf, partition_dict)
             self.dict_merge(partition_conf, f5_config_dict)
             f5_config_dict = partition_conf
         # Added not supported parse config to file
         merged_not_supported_list = not_supported_list + \
-                                    not_supported_list_partition
+            not_supported_list_partition
```
grastogi23: better to use () rather than +\

grastogi23 (approved):

PR #496, opened 2017-11-10
fixes the some missed clone pools [Ace]

Jeyanthinath:


python/avi/migrationtools/ace_converter/vs_converter.py
```diff
@@ -59,7 +59,7 @@ def create_http_policy(self, action, name):
     def check_persistance(self, pool_name, data, l4_type=None):
         for index, pool in enumerate(data['Pool']):
             if pool['name'] == pool_name:
-                if pool.get('application_persistence_profile_ref', ''):
+                if pool.get('application_persistence_profile_ref', '') or l4_type:
```
grastogi23: why not use and like

if (pool['name'] == pool_name) and (pool.get('application_persistence_profile_ref', '') or l4_type) and l4_type): 

Also, if the final condition is only based on l4_type then above logic is not right.

grastogi23 (changes requested):

python/avi/migrationtools/ace_converter/vs_converter.py
Jeyanthinath: fixed that , we are returning True in both that cases , also if l4 is there with other than client based ip address we are additionally removing it

python/avi/migrationtools/ace_converter/vs_converter.py
```diff
@@ -58,26 +58,26 @@ def create_http_policy(self, action, name):
...
-                        ref_name = pool['application_persistence_profile_ref'].split(
-                            '=')[-1]
-                        for app_name in data['ApplicationPersistenceProfile']:
-                            if app_name.get('name') == ref_name:
-                                if app_name.get('persistence_type') != 'PERSISTENCE_TYPE_CLIENT_IP_ADDRESS':
-                                    data['Pool'][index]['application_persistence_profile_ref'] = ''
-                                    LOG.warning("Removing persistance in Pool because of l4 type"
-                                                "cannot support cookie based persistance")
-                    return True
+            if pool['name'] == pool_name and (pool.get('application_persistence_profile_ref', '') or l4_type):
```
grastogi23, line 61: seems like the right logic is 
if pool['name'] == pool_name and pool.get('application_persistence_profile_ref') and l4_type

because next if will only get executed if l4_type is true.

grastogi23 (changes requested):

python/avi/migrationtools/ace_converter/vs_converter.py
Jeyanthinath: we are returning true on if any one of the "pool.get('application_persistence_profile_ref')"  or "l4_type" is present. 

The nested l4 if loop is to check and removing the application persistence over here

grastogi23 (approved):

PR #501, opened 2017-11-16
[Av 29290] Ansible: Add support to accept and return auth credentials in every module

chaitanyaavi:


python/avi/sdk/avi_api.py
```diff
@@ -161,6 +162,8 @@ def update_from_ansible_module(self, module):
             self.tenant = module.params['tenant']
         if module.params['tenant_uuid']:
             self.tenant_uuid = module.params['tenant_uuid']
+        if module.params['session_id']:
+            self.session_id = module.params['session_id']
```
grastogi23: is session id enough... I though you also need csrf token

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -29,7 +29,8 @@ def json(self):
         return self.obj
 
 
-def ansible_return(module, rsp, changed, req=None, existing_obj=None):
+def ansible_return(module, rsp, changed, req=None, existing_obj=None,
+                   api_context=None, session_id=None):
```
grastogi23: why need both. can we not keep everything inside api_context. also, let us not overload the keystone token with the csrf token. so we can create a new named tuple or a dict class that has 

'csrftoken' and 'session_id' as its members.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -427,10 +439,14 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
             rsp = api.post(obj_type, data=obj, tenant=tenant,
                            tenant_uuid=tenant_uuid, api_version=api_version)
     if rsp is None:
-        return module.exit_json(changed=changed, obj=existing_obj)
+        return module.exit_json(changed=changed, obj=existing_obj,
```
grastogi23, line 447: let us implement function call api.get_context() instead of enumerating the fields.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -446,4 +462,6 @@ def avi_common_argument_spec():
             tenant=dict(default='admin'),
             tenant_uuid=dict(default=''),
             api_version=dict(default='16.4.4', type='str'),
-            avi_credentials=dict(default=None, no_log=True, type='dict'))
+            avi_credentials=dict(default=None, no_log=True, type='dict'),
+            api_token=dict(type='str'),
```
grastogi23: let us only have api_context as the dict() and not two fields as api_token and session_id

grastogi23 (changes requested):

grastogi23 (approved):

PR #521, opened 2017-12-15
Added check for 404 status code

Rishabh-Baranwal:


go/session/avisession.go
```diff
@@ -281,7 +281,24 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
...
+			emsg := fmt.Sprintf("%v", mresult)
+			errorResult.Message = &emsg
+		}
+		return result, errorResult
+	}
+
 	result, err = ioutil.ReadAll(resp.Body)
+
+	if err != nil {
+		result = nil
```
grastogi23: Here we should still return errorResult. update the errorResult.Message to be err's message

grastogi23 (changes requested):

go/session/avisession.go
Rishabh-Baranwal: if err != nil {
		result = nil
		errorResult = AviError{verb: verb, url: url}
		errmsg := fmt.Sprintf("Response body read failed: %v", err)
		errorResult.Message = &errmsg
	}

	return result, errorResult

Tried with above but still failing with Encountered an error on POST request to URL https://<controller-ip>/login: HTTP code: 200.

go/session/avisession.go
```diff
@@ -281,7 +281,27 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 		return result, nil
 	}
 
+	if resp.StatusCode == 404 {
```
grastogi23: can we combine this code with line 267. It seems like this will never hit due to  	if resp.StatusCode < 200 || resp.StatusCode > 299 {

grastogi23 (changes requested):

go/session/avisession.go
grastogi23: ok, In this code we only need to fix the issue of returning errorResult properly when the parsing fails. no 404 handling here. handle the 404 inside the terraform code which is invoking the delete. That code should extract errorResult.http_status_code and ignore if it is 404.

grastogi23 (changes requested):

grastogi23 (approved):

ypraveen (approved):

PR #522, opened 2017-12-18
AV-31637 - Golang SDK Features & Bug Fixes

RahulMitra:
Golang SDK:
* Add support for token based authentication AviSession:
`
	authTokenSession, err := NewAviSession("localhost", "admin",
		SetAuthToken("auth-token-goes-here"),
		SetInsecure)
`
* Also add support for custom refresh token callback function which will generate auth token:
`
	authTokenSession, err := NewAviSession("localhost", "admin",
		SetRefreshAuthTokenCallback(getAuthToken),
		SetInsecure)
`
* API versioning, fixed to AviSession
`	authTokenSession, err = NewAviSession("localhost", "admin", SetPassword("password"),
		SetVersion("17.1.2"), SetInsecure)
`
* Bug fixes (fix some missing instantiations that caused code to not compile...)
* Update unit tests for both credentials and auth token session types


go/session/avisession.go
```diff
@@ -114,8 +127,37 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
...
+		return avisess, err
+	} else {
+		err := avisess.initiateSession()
+		return avisess, err
+	}
+}
+
+// In future, this will become an internal API
+func (avisess *AviSession) setAuthToken() error {
+	output, err := exec.Command("/opt/avi/python/bin/portal/manage.py", "gen_auth_token", "--user", "admin", "--hours", "12").Output()
```
grastogi23: this should not kept in the SDK as SDK is expected to run outside controller too. Rather have an api that takes argument of token. This code should be part of the internal AVI controller code and not exposed out.

go/session/avisession.go
```diff
@@ -158,6 +200,17 @@ func (avisess *AviSession) setPassword(password string) error {
 	return nil
 }
 
+func SetVersion(version string) func(*AviSession) error {
+	return func(sess *AviSession) error {
+		return sess.setTenant(version)
```
grastogi23: this should be setVersion rather than setTenant

go/session/avisession.go
```diff
@@ -76,13 +77,19 @@ type AviSession struct {
 	// Avi Controller.
 	password string
 
+	// tokenAuth specifies whether we should use an authenticatoin token
+	tokenAuth bool
```
grastogi23: we have two fields with confusing names tokenAuth and authToken. can we change this one to useTokenAuth. and also keep it together with the authtoken param.

go/session/avisession.go
```diff
@@ -264,6 +325,12 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 		return avisess.restRequest(verb, uri, payload)
 	}
 
+	if resp.StatusCode == 401 && avisess.tokenAuth {
```
grastogi23: again, this code should be outside the avisess sdk. may be we allow callers to register a callback function to refresh token. And then the caller can supply the new token everytime there is 401 and use token is set.

go/session/session_test.go
```diff
@@ -2,43 +2,59 @@ package session
...
+	"github.com/golang/glog"
 )
 
-func TestAviSession(t *testing.T) {
-	avisess, err := NewAviSession("10.10.25.201", "admin", SetPassword("avi123"), SetInsecure)
+func getSessions(t *testing.T) (
+	credentialsSession *AviSession, authTokenSession *AviSession) {
+	/* Test username/password authentication */
+	credentialsSession, err := NewAviSession("10.10.25.201",
+		"admin", SetPassword("avi123"), SetInsecure)
```
grastogi23: though it is old code. pl. change password to something other than avi123

go/session/session_test.go
```diff
@@ -57,44 +73,48 @@ func TestAviSession(t *testing.T) {
...
 }
 
-func TestAviPool(t *testing.T) {
-	avisess, err := NewAviSession("10.10.25.201", "admin", SetPassword("avi123"), SetInsecure)
-	if err != nil {
-		t.Error("Session Creation failed: ", err)
-		return
-	}
-
+func testAviPool(t *testing.T, avisess *AviSession) {
```
grastogi23: let us keep it consistent whether to have every test start with Caps or not. I don't know what is the Go Preferred way

grastogi23 (changes requested):

go/clients/avi_client.go
```diff
@@ -37,7 +37,9 @@ type AviClient struct {
 	DebugServiceEngine             *DebugServiceEngineClient
 	DebugVirtualService            *DebugVirtualServiceClient
 	Gslb                           *GslbClient
+	GslbApplicationPersistenceProfile *GslbApplicationPersistenceProfileClient
```
ypraveen, line 40: AviClient changes should be done in a separate commit, as they are version specific.

go/session/avisession.go
```diff
@@ -76,13 +77,19 @@ type AviSession struct {
 	// Avi Controller.
 	password string
 
+	// tokenAuth specifies whether we should use an authenticatoin token
```
ypraveen: typo in authentication

go/clients/avi_client.go
RahulMitra: Praveen, when I pulled the code from master, it would not compile without these additional changes. This is because the NewAviClient function tries to assign this field, which prior did not exist: 	

> aviClient.GslbApplicationPersistenceProfile = NewGslbApplicationPersistenceProfileClient(aviSession)

go/session/session_test.go
RahulMitra: This is actually the correct way for go. When structuring go unit tests, the tests which are run must have the format Test{FunctionName}. This is a decomposed helper function that is called by TestAviPool, the top-level function, so it must be lower case.

go/session/avisession.go
```diff
@@ -264,6 +331,19 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 		return avisess.restRequest(verb, uri, payload)
 	}
 
+	if resp.StatusCode == 401 && avisess.isTokenAuth() {
```
grastogi23: if session expired then would it always take the line 328 branch? all of them would be true in that case?

go/session/avisession.go
```diff
@@ -264,6 +331,19 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
...
+		// auth token incorrect; If callback function is provided,
+		// get new auth token and retry the request
+		if avisess.refreshAuthToken != nil {
+			avisess.setAuthToken(avisess.refreshAuthToken())
+			return avisess.restRequest(verb, uri, payload)
+		} else {
+			emsg := "Invalid Auth Token"
```
grastogi23: We should preserve the message from the resp rather than masking it with "Invalid auth token". This would be helpful in case of understanding why did controller reject and any additional clues it provided. Additionally, you can add no refresh token function supplied.

grastogi23 (changes requested):

go/session/avisession.go
```diff
@@ -264,6 +331,19 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 		return avisess.restRequest(verb, uri, payload)
 	}
 
+	if resp.StatusCode == 401 && avisess.isTokenAuth() {
+		// auth token incorrect; If callback function is provided,
+		// get new auth token and retry the request
+		if avisess.refreshAuthToken != nil {
+			avisess.setAuthToken(avisess.refreshAuthToken())
+			return avisess.restRequest(verb, uri, payload)
```
vivek-avi: if the token authentication fails constantly, could this become an infinite loop ?

vivek-avi: same issue can happen with the initiateSession() logic above, it needs to check the error returned by initiateSession and return on failure.

go/session/avisession.go
RahulMitra: I believe this to be the case. This is because the session ID has been set, and is not hitting the login URL, yet we got a 401. I believe this check is necessary because initiateSession itself makes a POST to the /login route, which therefore calls this function, so we don't want to get stuck in an infinite loop.

go/session/avisession.go
RahulMitra: Implemented a counting mechanism and some crude backoff for retrying to avoid infinite loop.

grastogi23 (approved):

PR #525, opened 2017-12-20
AV-31637 - Golang SDK Features & Bug Fixes (#522)

RahulMitra:
* Golang SDK: Add support for token based authentication AviSession; API versioning fixed to AviSession; Bug fixes

* Remove token auth functionality inside SDK and allow user to provide custom callback function

* Allow user to specify just a token callback function to start a session

* Add mechanism to retry 3 times and then stop, to avoid infinite loop


RahulMitra:
Cherry-pick from master onto 17.2.5

PR #526, opened 2017-12-20
removed repeated code and make purging non data args programmatic

grastogi23:


grastogi23:
api_context was getting included in the data comparison.

PR #530, opened 2018-01-02
Inventory for f5

Rishabh-Baranwal:


grastogi23:
@Jeyanthinath @Rishabh-Baranwal  let us rebase it to 17.2.8

Jeyanthinath:
@grastogi23 rebased, But please don't merge it now

python/avi/migrationtools/setup.py
```diff
@@ -33,7 +34,7 @@
                       'pycrypto', 'ecdsa', 'pyOpenssl', 'nose-html-reporting',
                       'nose-testconfig', 'ConfigParser', 'xlsxwriter', 'jinja2',
                       'pandas', 'openpyxl', 'appdirs',  'pexpect',
-                      'unittest2', 'networkx'],
+                      'unittest2', 'networkx', 'f5-sdk', 'bigsuds'],
```
grastogi23: we cannot have these tools be installed by avi. We need to raise proper error if f5-sdk and bigsuds is not found.

grastogi23 (changes requested):
let us change the dependency on f5-sdk and bigsuds

python/avi/migrationtools/setup.py
Jeyanthinath: so we need to remove the dependency over here and through exceptions only at run time if the packages or not found , isn't it ?

grastogi23:
Yes. Also this way those libraries would not get packed with controller and
customers will install them.

Gaurav Rastogi
Avi Networks


On Feb 21, 2018, at 11:58 AM, Jeyanthinath MuthuRam <
notifications@github.com> wrote:

*@Jeyanthinath* commented on this pull request.
------------------------------

In python/avi/migrationtools/setup.py
<https://github.com/avinetworks/sdk/pull/530#discussion_r169759928>:

> @@ -33,7 +34,7 @@
                       'pycrypto', 'ecdsa', 'pyOpenssl', 'nose-html-reporting',
                       'nose-testconfig', 'ConfigParser',
'xlsxwriter', 'jinja2',
                       'pandas', 'openpyxl', 'appdirs',  'pexpect',
-                      'unittest2', 'networkx'],
+                      'unittest2', 'networkx', 'f5-sdk', 'bigsuds'],

so we need to through exceptions at run time if the packages or not found ,
isn't it ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/avinetworks/sdk/pull/530#discussion_r169759928>, or mute
the thread

python/avi/migrationtools/f5_inventory.py
```diff
@@ -0,0 +1,508 @@
...
+import os
+import argparse
+import json
+import yaml
+import xlsxwriter
+try:
+    from bigsuds import BIGIP               # version 10
+    from f5.bigip import ManagementRoot     # version 11
+except ImportError:
+    print "Please Install bigsuds and f5 sdk's to proceeed.."
```
grastogi23: We should Log error as print may not work. Also, don't continue after this point. Simply raise the exception again.

Jeyanthinath: ok will exit the code after raising the exception

grastogi23:
@Rishabh-Baranwal @Jeyanthinath do we still keep this out?

python/avi/migrationtools/setup.py
grastogi23: this is back... we should remove 'f5-sdk', 'bigsuds' dependencies

grastogi23 (changes requested):

Jeyanthinath:
removed the package dependency from setup.py

grastogi23 (approved):

PR #534, opened 2018-01-09
AV-32390 : Avi api context changes for ansible utils 

chaitanyaavi:


grastogi23:
we need to wait till the 17.2.5

chaitanyaavi:
Blocked on https://avinetworks.atlassian.net/browse/AV-33013

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -30,25 +31,46 @@ def json(self):
...
+    api_creds.update_from_ansible_module(module)
+    key = '%s:%s:%s' % (api_creds.controller, api_creds.username,
+                        api_creds.port)
+    disable_fact = module.params['avi_disable_session_cache_as_fact']
+
+    fact_context = None
+    if not disable_fact:
+        fact_context = module.params.get('avi_api_context', {})
+        if fact_context:
+            fact_context.update({key: avi_api_context})
```
grastogi23: you can simply do 
fact_context[key] = avi_api_context
as it is always guarantee that fact_context will be {} or real context. Unless you believe it could be set with None.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -30,25 +31,46 @@ def json(self):
...
+        else:
+            fact_context = {key: avi_api_context}
+
+    obj_val = rsp.json() if rsp else existing_obj
+    old_obj_val = existing_obj if changed and existing_obj else None
+    avi_api_context_val = avi_api_context if disable_fact else None
+    ansible_facts_val = dict(
+        avi_api_context=fact_context) if not disable_fact else {}
```
grastogi23: why else {}. The default is probably None

grastogi23 (approved):

PR #539, opened 2018-01-19
#AV-33394: Added flag in sdk for lazy authentication of a session

chaitanyaavi:


grastogi23 (approved):

grastogi23:
AV-33355 #commit

nmandeyam (commented):
Lazy authentication should be false by default right?  The expected behavior is that when Api session is called without explicit parameters, should'nt it trigger a login?  because lazy_authentication is disabled?

PR #541, opened 2018-01-19
Changed lazy session to be false by default. AV-33355 #commit

grastogi23:


nmandeyam (approved):
Do we need to fix/add test-cases to ensure both the default and non-default behavior is exercised?

PR #542, opened 2018-01-22
Asible util minor fixes

chaitanyaavi:


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -490,17 +491,9 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
-                       avi_api_context=api.get_context())
-        # return module.exit_json(changed=changed, obj=existing_obj,
-        #                         ansible_facts=dict(avi_api_context=api.get_context()))
-    else:
-        return ansible_return(module, rsp, changed, req,
-                              existing_obj=existing_obj,
-                              avi_api_context=api.get_context()
-                              )
+
+    ansible_return(module, rsp, changed, req, existing_obj=existing_obj,
```
grastogi23: are you sure about this one. with this change it will not return the ansible facts?

grastogi23: also here it should be return ansible return

grastogi23 (changes requested):
this code needs more vetting.

python/avi/sdk/utils/ansible_utils.py
```diff
@@ -44,15 +44,16 @@ def ansible_return(module, rsp, changed, req=None, existing_obj=None,
     changed
     Returns: specific ansible module exit function
     """
-    if rsp and rsp.status_code > 299:
+
+    if  rsp != None and rsp.status_code > 299:
```
grastogi23: rsp is not None

grastogi23 (approved):

PR #554, opened 2018-01-31
AV-30240 #clearing session cookies before login

skdas:


grastogi23:
let us add test for it in the test_avi_api.py 
@patilshrikant486

PR #566, opened 2018-02-09
#AV-34084: SDK: Provide API to check if the connection to controller …

chaitanyaavi:
…was successful.


python/avi/sdk/avi_api.py
```diff
@@ -637,6 +650,7 @@ def get_object_by_name(self, path, name, tenant='', tenant_uuid='',
                         timeout=timeout,
                         params=params, api_version=api_version, **kwargs)
         if resp.status_code in (401, 419):
+            sessionDict[self.key]['connected'] = False
```
grastogi23: why not put this in the reset_session? so there is one place to flip the state.

grastogi23 (changes requested):

grastogi23 (approved):

grastogi23 (approved):

PR #569, opened 2018-02-15
Changed the behavior to match tenant_ref of object in case it is case…

grastogi23:
… of system profile

### DIFF
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -410,7 +410,7 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
 
         # Need to check if tenant_ref was provided and the object returned
         # is actually in admin tenant.
-        if 'tenant_ref' in obj and 'tenant_ref' in existing_obj:
+        if existing_obj and 'tenant_ref' in obj and 'tenant_ref' in existing_obj:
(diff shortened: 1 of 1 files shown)


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -407,6 +407,15 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
         existing_obj = api.get_object_by_name(
             obj_type, name, tenant=tenant, tenant_uuid=tenant_uuid,
             params=params, api_version=api_version)
+
+        # Need to check if tenant_ref was provided and the object returned
+        # is actually in admin tenant.
+        if 'tenant_ref' in obj and 'tenant_ref' in existing_obj:
```
chaitanyaavi: Existing object can be None in the case of create it fails in that scenario
works with following condition
if 'tenant_ref' in obj and existing_obj and 'tenant_ref' in existing_obj:

chaitanyaavi:
After changing the condition for create scenario tasks works as expected only if we pass the tenant in ref as well as with tenant param for login. Eg.

tenant_ref: /api/tenant/?name=foo
tenant: foo

chaitanyaavi (approved):

grastogi23:
AV-34843 #commit

PR #570, opened 2018-02-19
#AV-34518 : SDK get_object_by_name swallows invalid version errors - …

chaitanyaavi:
…Added logger staments

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -655,10 +655,14 @@ class ApiSession(Session):
                     path, name, tenant, tenant_uuid, timeout=timeout,
                     params=params, **kwargs)
         if resp.status_code > 299:
+            logger.error('Error in get object by name for %s named %s. '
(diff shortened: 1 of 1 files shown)


grastogi23 (changes requested):
I believe the ask is to throw exception in this case not just print errors.

python/avi/sdk/avi_api.py
```diff
@@ -655,10 +655,14 @@ def get_object_by_name(self, path, name, tenant='', tenant_uuid='',
                     path, name, tenant, tenant_uuid, timeout=timeout,
                     params=params, **kwargs)
         if resp.status_code > 299:
-            return obj
+            logger.error('Error in get object by name for %s named %s. '
```
grastogi23: no we should throw exception for anything that is not 404

grastogi23 (changes requested):
change needed as per discussion today

grastogi23 (approved):

PR #574, opened 2018-02-21
Added skip default flag to api options param

grastogi23:


ypraveen:
@grastogi23 Can you add "gofmt -l" to the travis job and make sure that it passes without errors?

ypraveen (approved):

ypraveen:
@grastogi23 we should remove the "Avi Confidential" copyright notice from the go files. We may need to add some notice that allows customers to freely use it but not sue us if something goes wrong.

PR #577, opened 2018-02-23
#AV-34384 : Added retries for authenticate session

chaitanyaavi:


grastogi23 (approved):

python/avi/sdk/avi_api.py
```diff
@@ -438,9 +439,23 @@ def authenticate_session(self):
...
+                raise
+            logger.warning('Connection error retrying %s', e)
+            time.sleep(5)
+            self.num_session_retries += 1
+            if self.num_session_retries > self.MAX_API_RETRIES:
+                self.num_session_retries = 0
+                raise APIError(
+                    "giving up after %d retries connection failure %s" %
+                    (self.MAX_API_RETRIES, True))
+            self.authenticate_session()
```
ypraveen, line 456: shouldn't we return after this? if not, then we will execute the following lines at all recursion levels.

Has this code been tested? May be mock and test?

PR #584, opened 2018-03-06
refactored the code around retry and wait on authentication

grastogi23:


python/avi/sdk/avi_api.py
```diff
@@ -230,6 +230,8 @@ def __init__(self, controller_ip=None, username=None, password=None,
         self.user_hdrs = {}
         self.data_log = data_log
         self.num_session_retries = 0
+        self.retry_wait_time = 0
```
nmandeyam, line 233: Suggest promoting them as 1st class citizens in the constructor api.  Since SDK users may want to modify this behaviour in an explicit manner

python/avi/sdk/avi_api.py
```diff
@@ -230,6 +230,8 @@ def __init__(self, controller_ip=None, username=None, password=None,
...
+        self.max_session_retries = self.MAX_API_RETRIES
```
nmandeyam, line 234: Please commit this fix into 17.2.8 as well as Eng-top.  I will test it on 17.2.8 and provide an update.

nmandeyam (commented):
I will have to tweak the gslb_site_housekeeping.txt suite to address the new timing behavior.   The verification window
will have to change with this commit.

python/avi/sdk/avi_api.py
```diff
@@ -442,40 +444,38 @@ def authenticate_session(self):
...
+                        'api': self,
+                        'connected': True
+                    }
+                logger.debug("authentication success for user %s",
+                             self.avi_credentials.username)
+                return
         except (ConnectionError, SSLError) as e:
             if not self.retry_conxn_errors:
                 raise
             logger.warning('Connection error retrying %s', e)
```
ypraveen, line 467: should we catch other errors and raise right away?

ypraveen (approved):

PR #589, opened 2018-03-09
Changed the code to return the actual errors rather than overshadow them

grastogi23:


python/avi/sdk/avi_api.py
```diff
@@ -441,6 +441,7 @@ def authenticate_session(self):
             raise APIError("Neither user password or token provided")
```
nmandeyam, line 441: Why are we waiting in authenticate to log this error? Can't we make it mandatory check in the constructor path?

python/avi/sdk/avi_api.py
```diff
@@ -461,20 +462,26 @@ def authenticate_session(self):
                 logger.debug("authentication success for user %s",
                              self.avi_credentials.username)
                 return
+            else:
+                logger.error("Error status code %s msg %s", rsp.status_code,
+                             rsp.text)
+                err = APIError('Status Code %s msg %s' % (
```
nmandeyam, line 468: will rsp.txt and rsp.status_code always be present?  I vaguely remember in some scenarios, these fields may not be set in the rsp and application trying to reference these fields was causing exceptions.

nmandeyam, line 468: Do we need to do something similar like this in the lazy authentication path?

def _api():
<snip>

except (ConnectionError, SSLError) as e:
            logger.warning('Connection error retrying %s', e)
            if not self.retry_conxn_errors:
                raise
            connection_error = True
        except Exception as e:
            logger.error('Error in Requests library %s', e)
            raise
        if not connection_error:
            logger.debug('path: %s http_method: %s hdrs: %s params: '
                         '%s data: %s rsp: %s', fullpath, api_name.upper(),
                         api_hdrs, kwargs, data,
                         (resp.text if self.data_log else 'None'))
        if connection_error or resp.status_code in (401, 419):
            if connection_error:
                try:
                    self.close()
                except:
                    # ignoring exception in cleanup path
                    pass
                logger.warning('Connection failed, retrying.')
                # Adding sleep before retrying
                if self.retry_wait_time:
                    time.sleep(self.retry_wait_time)
            else:
                logger.info('received error %d %s so resetting connection',
                            resp.status_code, resp.text)
            ApiSession.reset_session(self)
            self.num_session_retries += 1
            if self.num_session_retries > self.max_session_retries:
                # Added this such that any code which re-tries can succeed
                # eventually.
                self.num_session_retries = 0
                raise APIError(
                    "giving up after %d retries connection failure %s" %
                    (self.max_session_retries, connection_error))
            # should restore the updated_hdrs to one passed down

grastogi23: no

PR #645, opened 2018-05-22
AV-39668: Test case for vrf ref

patilshrikant486:


python/avi/migrationtools/test/common/config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-controller_version_v16: 16.4.4
-controller_version_v17: 17.2.1
+controller_version_v16: 16.5.4
+controller_version_v17: 17.2.10
```
chaitanyaavi: Update to 17.2.12

patilshrikant486: Updated controller version 17.2.10 to 17.2.12

grastogi23 (approved):

PR #646, opened 2018-05-23
Add cloning between controllers. Support more complex configurations.…

thebertster:
Code rewritten to be more modular and extensible. Now handles cloning to a different controller and more complex configurations. Also added options to force cloning of certain child objects rather than re-using them (e.g. application profile, health monitors etc.). Miscellaneous other enhancements.


python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
 
-        if old_generic_name.startswith(object_type + '/'):
-            g_obj = self.api.get(old_generic_name, tenant_uuid = t_obj['uuid'] if t_obj else None).json()
-            old_generic_name = g_obj['name']
+        force_clone = force_clone or []
+
+        # Retrieve tenant, other_tenant and other_cloud names, uuids and objects
+        # which may have been passed as objects or names
+
+        if t_obj:
```
grastogi23: let us break this into smaller sections such that it can be better maintained in future.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
+            else:
+                t_obj = self.api.get_object_by_name('tenant', tenant)
+                if t_obj is None:
+                    raise Exception('A tenant with name %s could not be found'
+                                    % tenant)
+                tenant_uuid = t_obj['uuid']
+
+        if ot_obj:
+            other_tenant = ot_obj['name']
+            otenant_uuid = ot_obj['uuid']
```
grastogi23: seems like this can be separted out into another method that is invoked for t_obj and ot_obj

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
         f = kwargs.get('file', sys.stdout)
         f.flush() if f is not None else sys.stdout.flush()
 
-
 class AviClone:
-    def __init__(self, api):
-        self.api = api
+    VALID_POOL_REF_OBJECTS = {
```
grastogi23, line 42: we may need to add pool group support too. Specially verizon is using pool groups.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
+        api = api_to_use or self.api
+
         obj = None
         if not params:
             params = {}
         params['name'] = name
-        resp = self.api.get(path, tenant, tenant_uuid, timeout=timeout,
+        resp = api.get(path, tenant, tenant_uuid, timeout=timeout,
                         params=params, api_version=api_version, **kwargs)
         if resp.status_code in (401, 419):
```
grastogi23, line 101: sdk is already doing these you don't need to do this again.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
-                    params=params, **kwargs)
+            resp = api.get_object_by_name(path, name, tenant, tenant_uuid,
+                                          timeout=timeout, params=params,
+                                          **kwargs)
         if resp.status_code < 300:
             obj = resp.json()['results']
 
-        self.api._update_session_last_used()
-        
+        api._update_session_last_used()
```
grastogi23: same here. sdk is already doing it internally.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
+                    raise Exception('A cloud with name %s could not be found'
+                                    % other_cloud)
+
+        if not object_type:
+            # If object_type is not specified, assume the old_name is in
+            # form object_type/uuid
+
+            object_type = old_name.split('/')[0]
+
+        logger.debug('Cloning %s "%s" to "%s"', object_type,
```
grastogi23, line 241: split code to be in two lines.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -36,354 +37,406 @@ def spprint(s, ind='', **kwargs):
...
+                                       new_name, r.status_code, r.text))
+                logger.debug(exception_string)
+                logger.debug(old_obj)
+                raise Exception(exception_string)
 
-    def clone_pool(self, old_pool_name, new_pool_name, tenant = None,
-                   other_tenant = None, other_cloud = None,
-                   force_unique_name = False):
+        except Exception as ex:
+            # If an exception occurred, delete any intermediate objects we
```
grastogi23: This would be important thing to implement. alternatively, you can collect all objects and use api/macro to create them in one go. Not sure if you aware of it. I can go over with you about intro to macro api.

grastogi23 (changes requested):

python/avi/sdk/samples/clone_vs.py
```diff
@@ -1225,24 +1931,85 @@ def clone_refs_to_tenant(self, parent_obj, refs, t_obj, ot_obj):
 VS, including any specified in context-switching rules, will also be cloned.
 
 The script also allows the cloning of objects into a different tenant and/or 
-cloud than the source. This may not always be successful because the source 
-and destination tenant/cloud may have different properties. If the clone attempt
-fails, the script will automatically delete any objects that it created along 
-the way.
+cloud than the source - the destination may also be on a different Avi Vantage
```
grastogi23: let us make this part of the help message for the script. It would be very useful in the field.

grastogi23 (changes requested):

python/avi/sdk/samples/clone_vs.py
thebertster: Cloning of Pool Groups is already catered for. This dictionary just lists a set of attributes which the user can specify to be forcibly cloned rather than re-used. I couldn't see any obvious pool group attributes that would need this treatment.

python/avi/sdk/samples/clone_vs.py
thebertster: get_all_objects_by_name() is simply a straight copy of the get_object_by_name() from the SDK but instead of returning only the first object matching the name, returns all objects. I think we discussed this last time we reviewed the code and there was a suggestion to add this function directly to the SDK as it would be useful elsewhere...

python/avi/sdk/samples/clone_vs.py
thebertster: get_all_objects_by_name() is simply a straight copy of the get_object_by_name() from the SDK but instead of returning only the first object matching the name, returns all objects. I think we discussed this last time we reviewed the code and there was a suggestion to add this function directly to the SDK as it would be useful elsewhere...

python/avi/sdk/samples/clone_vs.py
thebertster: This is already fully implemented.

python/avi/sdk/samples/clone_vs.py
thebertster: Could you clarify what you need to see here?

grastogi23: logger.debug('Cloning %s "%s" to "%s"', object_type,
    old_name, new_name)

python/avi/sdk/samples/clone_vs.py
```diff
@@ -1168,81 +1814,177 @@ def clone_refs_to_tenant(self, parent_obj, refs, t_obj, ot_obj):
...
- to fail as it may reference shared objects which do not make sense in the 
- destination cloud
-* Cloning a VS to a different tenant/cloud will try to find an SE group with the
- same name as referenced in the source VS but if no match is found, will place
-  into the Default SE group instead
-* The script has been primarily written for and tested with Linux/VMWare/AWS
- clouds - it may not work as-is for other clouds
-'''
\ No newline at end of file
+        # If run with no parameters, display general help text for the script
```
grastogi23: Pl. look at        https://github.com/avinetworks/sdk/blob/e142a13e0e0f3e4fadf61357d9c2f59acf8e606a/python/avi/migrationtools/f5_converter/f5_converter.py#L438

on how you can pass the help string to the arg parser
ideally don't want to be mixing logger and printf

thebertster: OK, will do. I did it that way first, but didn't like the way it wrapped the text.

grastogi23 (approved):

grastogi23:
done

PR #658, opened 2018-06-06
 AV-40311: Fix for lb algorithm miss-match

chaitanyaavi:


grastogi23:
@chaitanyaavi can we also add unit test cases for them. it is easy get injections in such corner cases.

grastogi23 (approved):

grastogi23 (approved):

PR #679, opened 2018-06-26
AV-36570 #comment Added support for fileservice in GO SDK for 18.1.3 …

gitshrikant:
AV-36570 #comment Added support for fileservice in GO SDK for 18.1.3 version.


go/session/avisession.go
```diff
@@ -249,38 +252,44 @@ func (avisess *AviSession) isTokenAuth() bool {
...
+		glog.Errorf("[ERROR] mustOpen Error while opening  file %v", f)
+		panic(err)
+	}
+	return r
+}
+
 // restRequest makes a REST request to the Avi Controller's REST API.
 // Returns a byte[] if successful
-func (avisess *AviSession) restRequest(verb string, uri string, payload interface{}, retryNum ...int) ([]byte, error) {
+func (avisess *AviSession) restRequest(verb string, uri string, payload interface{}) ([]byte, error) {
```
grastogi23: we don't want to remove the retry logic....

grastogi23 (changes requested):

go/session/avisession.go
```diff
@@ -388,6 +397,308 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
...
+	if avisess.tenant != "" {
+		req.Header.Set("X-Avi-Tenant", avisess.tenant)
+	}
+	if avisess.sessionid != "" {
+		req.AddCookie(&http.Cookie{Name: "sessionid", Value: avisess.sessionid})
+		req.AddCookie(&http.Cookie{Name: "avi-sessionid", Value: avisess.sessionid})
+	}
+
+	dump, err := httputil.DumpRequestOut(req, true)
+	debug(dump, err)
```
grastogi23: why debug... we should remove it. only do the debug in case of error not otherwise.

go/session/avisession.go
```diff
@@ -408,6 +719,38 @@ func debug(data []byte, err error) {
...
+	if err != nil{
+		glog.Errorf("CheckControllerStatus Error while creating http request.")
+		return false, err
+	}
+
+	state_resp, err := client.Do(check_req)
+
+	if state_resp == nil{
+		time.Sleep(3 * time.Second)
+		return avisess.CheckControllerStatus()
```
grastogi23: this logic is not clear. Is the goal to wait for controllerstatus? if yes, then should the GET be re-tried. If not then it should not return output off the controller status.

grastogi23 (changes requested):

grastogi23 (approved):

PR #682, opened 2018-06-27
AV-41513 #comment Added wait for controller to get ready.

gitshrikant:


go/session/avisession.go
```diff
@@ -251,36 +251,33 @@ func (avisess *AviSession) isTokenAuth() bool {
 
 // restRequest makes a REST request to the Avi Controller's REST API.
 // Returns a byte[] if successful
-func (avisess *AviSession) restRequest(verb string, uri string, payload interface{}, retryNum ...int) ([]byte, error) {
+func (avisess *AviSession) restRequest(verb string, uri string, payload interface{}) ([]byte, error) {
```
grastogi23, line 254: why are we removing retry?

grastogi23 (changes requested):

go/session/avisession.go
```diff
@@ -251,36 +251,33 @@ func (avisess *AviSession) isTokenAuth() bool {
...
-	if retry > 3 {
-		errorResult := AviError{verb: verb, url: url}
-		errorResult.err = fmt.Errorf("tried 3 times and failed")
-		return nil, errorResult
-	}
+	client := &http.Client{Transport: tr}
 
-	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: avisess.insecure},
+	//Checking controller state form login uri
```
grastogi23, line 265: let us also split this into its own method so this core function does not get so long.

grastogi23 (changes requested):

PR #698, opened 2018-07-19
#AV-42110: When app profile is undefined, use System-L4-Application p…

chaitanyaavi:
…rofile

Minor changes


grastogi23 (approved):

grastogi23:
@chaitanyaavi @patilshrikant486  please add test case

PR #702, opened 2018-07-24
AV-40567 #comment 'Object not found' exception is raised for 404 API response.

gitshrikant:
AV-40567 #comment 'Object not found' exception is raised for 404 API response. An error message is reported for the raised exception.

### DIFF
--- a/python/avi/sdk/avi_api.py
+++ b/python/avi/sdk/avi_api.py
@@ -100,7 +100,7 @@ class ApiResponse(Response):
             return None
         elif self.status_code == 404:
             raise ObjectNotFound('HTTP Error: %d Error Msg %s' % (
-                                    self.status_code, self.text))
+                                    self.status_code, self.text), self)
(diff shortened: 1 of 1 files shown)


python/avi/sdk/avi_api.py
```diff
@@ -99,7 +99,8 @@ def json(self):
             # No response needed; e.g., delete operation
             return None
         elif self.status_code == 404:
-            raise ObjectNotFound()
+            raise ObjectNotFound('HTTP Error: %d Error Msg %s' % (
```
grastogi23, line 102: Please also pass self (response) in the exception.
ObjectNotFound('HTTP Error: %d Error Msg %s' % (
+                                    self.status_code, self.text), self)

grastogi23 (changes requested):
return response as well.

grastogi23 (approved):

PR #723, opened 2018-08-10
AV-40602 #comment Added support for user CRUD operation. 

gitshrikant:
Added support for username and password fields for user module.

### DIFF
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -10,7 +10,6 @@ from copy import deepcopy
 from avi.sdk.avi_api import ApiSession, ObjectNotFound, avi_sdk_syslog_logger, \
     AviCredentials
 
-
 if os.environ.get('AVI_LOG_HANDLER', '') != 'syslog':
     log = logging.getLogger(__name__)
 else:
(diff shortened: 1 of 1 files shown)


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -366,11 +365,18 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
...
-    if uuid  and obj_type != 'cluster':
+    if uuid and obj_type != 'cluster':
         obj_path = '%s/%s' % (obj_type, uuid)
     else:
         obj_path = '%s/' % obj_type
     obj = deepcopy(module.params)
+    tenant = obj.pop('tenant', '')
+    tenant_uuid = obj.pop('tenant_uuid', '')
+    # obj.pop('cloud_ref', None)
+    for k in POP_FIELDS:
```
grastogi23, line 383: If you move this up then obj_username would always be popped so username will never be passed down. Did the test work?

grastogi23 (changes requested):

python/avi/sdk/utils/ansible_utils.py
gitshrikant: yes. The test did work. Earlier it was always poping username and password that's why I moved it up. Now it is not poping the username and password field and it does work.

grastogi23:
How about we only allow user module to use api_credential. then there will be no issues?

grastogi23 (approved):

grastogi23 (approved):

PR #731, opened 2018-08-31
Added test case for add cs action

patilshrikant786:


chaitanyaavi:
Add this to jenkins not travis

PR #744, opened 2018-09-24
AV-45356 Avoid retry for 4XX response codes.

gitshrikant:


grastogi23:
can you please add test case for this one

python/avi/sdk/avi_api.py
```diff
@@ -468,6 +468,14 @@ def authenticate_session(self):
                 logger.debug("authentication success for user %s",
                              self.avi_credentials.username)
                 return
+            # Check for bad request and invalid credentials response code
+            elif rsp.status_code >= 400 and rsp.status_code < 500:
```
grastogi23: let us check for exact error for the invalid credentials. I believe it is 401

grastogi23: you can also include 403 as well.

grastogi23 (changes requested):

grastogi23 (approved):

PR #750, opened 2018-10-03
AV-45206 Modified controller wait for response code 500. Cluster creation is successful after this change.

gitshrikant:


go/session/avisession.go
```diff
@@ -726,7 +726,7 @@ func (avisess *AviSession) CheckControllerStatus() (bool, error){
 		state_resp, err := avisess.client.Do(check_req)
 		if state_resp != nil {
 			//Checking controller response
-			if state_resp.StatusCode != 503 && state_resp.StatusCode != 502 {
+			if state_resp.StatusCode != 503 && state_resp.StatusCode != 502 && state_resp.StatusCode != 500 {
```
grastogi23, line 729: better way to code would be to say state_resp.StatusCode not in (503, 502, 500)

grastogi23 (changes requested):
see attached

go/session/avisession.go
gitshrikant: I think there is no built-in operator (like 'not in' ) to do it in Go. Otherwise, we need to iterate over the array.

grastogi23 (approved):

PR #772, opened 2018-10-31
#AV-47711: Issue with SSL Certs during migration fromNetScaler

chaitanyaavi:


python/avi/migrationtools/netscaler_converter/ns_constants.py
```diff
@@ -9,6 +9,7 @@
 STATUS_SKIPPED = 'SKIPPED'
 STATUS_SUCCESSFUL = 'SUCCESSFUL'
 STATUS_INDIRECT = 'INDIRECT'
+STATUS_DUMMY = 'DUMMY'
```
chaitanyaavi, line 12: @grastogi23 Let me know if this status for Excel need to be changed to something else.

grastogi23 (approved):

PR #791, opened 2018-11-21
#AV-51672: Changes for moving avi trafic from SDk to Avi Traffic Role

chaitanyaavi:


chaitanyaavi:
Cannot merge until role is available on Galaxy

grastogi23 (approved):

PR #792, opened 2018-11-22
AV-46644 num_api_retries added instead  num_session_retries in _api

chauhanshubham:


python/avi/sdk/avi_api.py
```diff
@@ -615,11 +616,11 @@ def _api(self, api_name, path, tenant, tenant_uuid, data=None,
                 logger.info('received error %d %s so resetting connection',
                             resp.status_code, resp.text)
             ApiSession.reset_session(self)
-            self.num_session_retries += 1
-            if self.num_session_retries > self.max_session_retries:
+            self.num_api_retries += 1
```
grastogi23, line 619: can we not simply use num_session_retries instead of num_api_retries?

grastogi23 (changes requested):
let us test that it actually works. based on the code it is not modifying the logic but simply changing the variable names.

python/avi/sdk/avi_api.py
chauhanshubham: Since `reset_session` resets the value of `num_session_retries` (which was causing the loop), there is, I think a need to track the retries by another counter, independent of `num_session_retries`.
I tested this by explicitly clearing the cookies every time we make an API request (in `def _api`) and this works well (stops retry attempts).
If we go ahead and use `num_session_retries`, we'll need a separate variable which forces it to NOT reset `num_session_retries` when a 401/419 is received in response to an API call. The current solution seemed more straightforward and simple.

grastogi23 (approved):

shubhamavi:
AV-46644

PR #806, opened 2018-12-10
AV-52543 Logout inactive sessions.

gitshrikant:


grastogi23 (approved):

grastogi23:
@gitshrikant  I don't see any test case change with this? It would be good to add a test case to try to check if the user really logs out.

grastogi23 (changes requested):
@gitshrikant I don't see any test case change with this? It would be good to add a test case to try to check if the user really logs out.

PR #814, opened 2018-12-14
GO SDK Spring Cleaning.

grastogi23:
Refactored Go code to remove code duplication and also missing functionality. Also, making resp.body.close() explicitly before doing retries as otherwise it forces Go to use a new TCP connection.

### DIFF
--- a/go/examples/api_reads.go
+++ b/go/examples/api_reads.go
@@ -4,8 +4,8 @@ import (
 	"github.com/avinetworks/sdk/go/session"
 	"github.com/golang/glog"
 	"os"
-	"time"
 	"strconv"
+	"time"
 )
 
 var AVI_CONTROLLER = os.Getenv("AVI_CONTROLLER")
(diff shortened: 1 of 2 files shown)


go/examples/api_reads.go
```diff
@@ -48,56 +48,80 @@ func main() {
...
+	if err != nil {
+		glog.Errorf("Login Failed settings CNTLR: %s USER: %s TENANT: %s VERSION: %s AUTH: %s ITR %d err %v",
+			AVI_CONTROLLER, AVI_USERNAME, AVI_TENANT, aviVersion, AVI_AUTH_TOKEN, AVI_API_ITERATIONS, err)
+		return
 	}
 
 	for i := 0; i < AVI_API_ITERATIONS; i++ {
 		start := time.Now()
 		var res interface{}
+		var err error
```
RahulMitra, line 72: Just from a code quality perspective, rather than defining `var err error` and assigning to that variable, I like to re-instantiate err each time:
`if err := <statement>; err != nil {`

RahulMitra (approved):
LGTM

gitshrikant (approved):

PR #826, opened 2018-12-21
allow controller as entity_type

ericsysmin:


python/avi/sdk/utils/api_utils.py
```diff
@@ -76,7 +76,10 @@ def get_metrics(
         if entity_name and not entity_uuid:
             resp = self.api.get_object_by_name(entity_type, entity_name)
             entity_uuid = self.api.get_obj_uuid(resp)
-        path = 'analytics/metrics/%s/%s' % (entity_type, entity_uuid)
+        if entity_type == 'controller':
```
grastogi23, line 76: this may not be enough.. the line 77 may fail. can you try to see it works? also commit it to 18.2.2 branch so it gets bundled with the controller as well.
thanks
Gaurav

grastogi23 (changes requested):
pl. see the comments.

python/avi/sdk/utils/api_utils.py
```diff
@@ -76,7 +76,13 @@ def get_metrics(
...
```
ericsysmin: @grastogi23 I think this may fix it it did fail this looks like it works however not sure if best way to do this

python/avi/sdk/utils/api_utils.py
ericsysmin: Updated, it did fail when I tested

ericsysmin:
@grastogi23 tested, works now, because there's no `entity_name` it was tossing an error. Once you make that an optional value, everything works well.

grastogi23 (approved):

PR #835, opened 2019-01-11
AV-53736: Go SDK enhancements

chauhanshubham:


grastogi23:
@shubhamavi  please add an example of converting error to AviError and access the information in side it. You can do simple stuff like bad post or login etc.

grastogi23 (approved):

chauhanshubham:
@grastogi23 added an example test to convert errors into instance of AviError and accessing AviError fields - HttpStatusCode, Url, Verb etc.

PR #849, opened 2019-02-15
AV-55613 AV-55707 AV-55706 AV-55704 saml sdk epic 18.2.2

gitshrikant:
Following is the example code to do SAML authentication for given IDPs and get the controller session.

Eg.
```
from avi.sdk.saml_avi_api import OktaSAMLApiSession
# default controller session
s = ApiSession("controller_ip", username="foo", password="foo")

# SAML based controller session for Okta IDP
s = ApiSession.get_session("controller_ip", username="foo", password="foo", idp=OktaSAMLApiSession)

# SAML based controller session for Okta IDP
s = OktaSAMLSession("controller_ip", username="foo", password="foo")

# SAML based controller session for Onelogin IDP
s = OneloginSAMLSession("controller_ip", username="foo", password="foo")

rsp = s.get('healthmonitor')
print rsp.text
rsp = s.get('pool')
print rsp.text
s.close()
```


python/avi/sdk/avi_api.py
```diff
@@ -949,6 +949,18 @@ def _clean_inactive_sessions():
                 continue
             keys_to_delete.append(key)
         for key in keys_to_delete:
+            # Logout inactive sessions
+            try:
+                key_split = key.split(':')
+                inactive_session = ApiSession.get_session(
```
grastogi23: why do we need to fetch session again?

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,722 @@
+from avi.sdk.avi_api import *
```
grastogi23: bad practice of importing *

grastogi23 (commented):
pl. review.

python/avi/sdk/avi_api.py
```diff
@@ -949,6 +949,18 @@ def _clean_inactive_sessions():
...
+                    controller_ip=key_split[0],
+                    username=key_split[1],
+                    port=key_split[2],
+                    csrftoken=session_cache[key]['csrftoken'],
+                    session_id=session_cache[key]['session_id'])
+                inactive_session.post('logout')
+            except Exception as e:
```
grastogi23: no error when sessio nis not found.

python/avi/sdk/saml_avi_api.py
gitshrikant: Removed import * and importing only required packages.

python/avi/sdk/avi_api.py
gitshrikant: Refactorred code and using existing inactive session to call logout URI.

python/avi/sdk/avi_api.py
gitshrikant: Handled exception and pass this exception just by reporting warning "Session Not Found."

gitshrikant (commented):
Abstraction of SAMLApiSession class is been implemented because to integrate this SAML SDK into the Ansible we will not need to mention specific IDP class while creating the session. Using this abstract class, a user just provides the IDP input and this class will take care of creating the specific IDP class object.

python/avi/sdk/avi_api.py
```diff
@@ -282,8 +287,19 @@ def __init__(self, controller_ip=None, username=None, password=None,
...
+            if idp:
+                idp_class = ApiSession.get_idp_class(idp)
+                idp_class(controller_ip, username, password, token=token,
+                          tenant=tenant, tenant_uuid=tenant_uuid,
+                          verify=verify, port=port, timeout=timeout,
+                          retry_conxn_errors=retry_conxn_errors,
+                          api_version=api_version, data_log=data_log,
+                          avi_credentials=avi_credentials,
+                          lazy_authentication=lazy_authentication,
+                          max_api_retries=max_api_retries, idp=idp)
```
grastogi23: why do we need the idp class here.  The idpclass should have their own derived class implementations of the self.authenticate_session()

python/avi/sdk/avi_api.py
```diff
@@ -366,24 +382,24 @@ def session_id(self):
 
     def get_context(self):
         return {
-            'session_id':sessionDict[self.key]['session_id'],
-            'csrftoken': sessionDict[self.key]['csrftoken']
+            'session_id':
+                sessionDict[self.key]['session_id'],
```
grastogi23: why split in two lines?

python/avi/sdk/avi_api.py
```diff
@@ -412,21 +429,26 @@ def get_session(
         key = '%s:%s:%s' % (avi_credentials.controller,
                             avi_credentials.username, k_port)
         cached_session = sessionDict.get(key)
+        idp_class = ApiSession
```
grastogi23: let the get_idp_class return ApiSession if idp is null.

python/avi/sdk/avi_api.py
```diff
@@ -412,21 +429,26 @@ def get_session(
...
+                controller_ip, username, password, token=token,
+                tenant=tenant, tenant_uuid=tenant_uuid,
+                verify=verify, port=port, timeout=timeout,
+                retry_conxn_errors=retry_conxn_errors,
                 api_version=api_version, data_log=data_log,
                 avi_credentials=avi_credentials,
                 lazy_authentication=lazy_authentication,
-                max_api_retries=max_api_retries)
-            ApiSession._clean_inactive_sessions()
+                max_api_retries=max_api_retries, idp=idp)
```
grastogi23: idp classes should remember the idp value. they should not take this parameter again.

python/avi/sdk/avi_api.py
```diff
@@ -442,6 +464,23 @@ def reset_session(self):
...
+            idp_class = OktaSAMLApiSession
+            logger.info("Using OktaSAMLApiSession to create "
+                        "controller session")
+        elif str(idp).lower() == "onelogin":
+            from avi.sdk.saml_avi_api import OneloginSAMLApiSession
+            idp_class = OneloginSAMLApiSession
+            logger.info("Using OneloginSAMLApiSession to create "
+                        "controller session")
+        else:
+            logger.error("Provided IDP implementation not found.")
```
grastogi23: It should simply return ApiSession class for none case which is default.

python/avi/sdk/avi_api.py
```diff
@@ -522,15 +561,19 @@ def _get_api_headers(self, tenant, tenant_uuid, timeout, headers,
             "Content-Type": "application/json"
         })
         api_hdrs['timeout'] = str(timeout)
-        if self.key in sessionDict and 'csrftoken' in sessionDict.get(self.key):
-            api_hdrs['X-CSRFToken'] = sessionDict.get(self.key)['csrftoken']
+        if self.key in sessionDict and 'csrftoken' in \
```
grastogi23, line 544: I don't like unnecessary use of "\\". It makes the code ugly.

python/avi/sdk/avi_api.py
```diff
@@ -522,15 +561,19 @@ def _get_api_headers(self, tenant, tenant_uuid, timeout, headers,
...
+                sessionDict.get(self.key):
+            api_hdrs['X-CSRFToken'] = \
+                sessionDict.get(self.key)['csrftoken']
         else:
             self.authenticate_session()
-            api_hdrs['X-CSRFToken'] = sessionDict.get(self.key)['csrftoken']
+            api_hdrs['X-CSRFToken'] = \
```
grastogi23: no need to change these lines.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,520 @@
...
+        try:
+            saml_controller_url = self.prefix + self.SAML_URL_SUFFIX
+            resp = controller_session.get(saml_controller_url,
+                                          allow_redirects=True)
+        except Exception as e:
+            logger.error("Error for controller SSO login URL. %s",
+                         str(e.message))
+            err = APIError('Status Code %s msg %s' % (
+                resp.status_code, resp.text), resp)
+            raise StandardError(err)
```
grastogi23: why do we need to convert to the StandardError instead of simply using ApiError? this is change in the intervace.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,520 @@
...
+            logger.info("Controller generated SAML request is being "
+                        "sent to IDP with existing IDP cookies.")
+            try:
+                resp = idp_session.post(assertion_url, headers=headers,
+                                        data=saml_data, allow_redirects=False,
+                                        cookies=self.idp_cookies)
+            except Exception as e:
+                logger.error("Error No SAML response from IDP %s",
+                             str(e.message))
+                err = APIError("Error No SAML response from IDP. "
```
grastogi23: can we be consistent on the ApiError vs standarderror.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,520 @@
...
+                                re.M | re.S).group(1)
+        assertion_url = re.search(assertion_url_regex, resp.text,
+                                  re.M | re.S).group(1)
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        saml_data = urllib.urlencode({
+            'SAMLRequest': saml_request,
+            'RelayState': relay_state})
+        if self.idp_cookies:
+            logger.info("Controller generated SAML request is being "
+                        "sent to IDP with existing IDP cookies.")
```
grastogi23: we should print out the controller IP for which SAML is being set and may be the SAML url as well,.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,520 @@
...
+                                                          resp.text), resp)
+                raise StandardError(err)
+        else:
+            try:
+                idp_resp = idp_session.post(assertion_url, headers=headers,
+                                            data=saml_data,
+                                            allow_redirects=False)
+            except Exception as e:
+                logger.error("Error No SAML response from IDP %s",
+                             str(e.message))
```
grastogi23: no need to do str()

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,520 @@
...
+                err = APIError("Error No SAML response from IDP. "
+                               "Status Code %s msg %s" % (resp.status_code,
+                                                          resp.text), resp)
+                raise StandardError(err)
+        else:
+            try:
+                idp_resp = idp_session.post(assertion_url, headers=headers,
+                                            data=saml_data,
+                                            allow_redirects=False)
```
grastogi23: what if there are no exception and response is simply 401 or 5xx?

grastogi23 (changes requested):
can you please cleanup the code and remove all the strs and redundant code. I am not convinced of all the error handling in the interactions during SAML assertions.

python/avi/sdk/avi_api.py
gitshrikant: I was following 80 characters pep8 standard. As this line exceeds 80 characters I need to split it.

python/avi/sdk/avi_api.py
gitshrikant: Unnecessary split. Reverting this.

python/avi/sdk/avi_api.py
gitshrikant: Changed get_idp_class to return ApiSession if idp is None.

python/avi/sdk/avi_api.py
gitshrikant: An IDP class has its own implementation of authenticate_session() method. IDP class just extends authenticate_session method from the base class AviSession. To override the base class's (AviSession) authenticate_method we required this initialization of idp_class in the AviSession class. This way self.authenticate_session method of Avisession class is overridden with the IDP class's authenticate_session method.

python/avi/sdk/saml_avi_api.py
gitshrikant: removed str()

python/avi/sdk/saml_avi_api.py
gitshrikant: Using ApiError instead of StandardError.

python/avi/sdk/saml_avi_api.py
gitshrikant: Printing SAML controller URL.

python/avi/sdk/saml_avi_api.py
gitshrikant: Added checks for response codes from IDP.

python/avi/sdk/avi_api.py
gitshrikant: The idp parameter is we need to specify when calling the get_session static method. This method is called without class instance. How can the IDP class remember this?

gitshrikant (commented):
@grastogi23 Travis is failing because of the coverage test. SAML tests are added and they can not be run on Travis. So, they are skipped and test coverage drop below 80% which is a threshold to pass minimum coverage required to pass the Travis. Should we set this threshold to 75% or is there any other way?

README.md
```diff
@@ -84,6 +84,64 @@ user=os.environ.get('USER')
...
+Currently, SDK support two IDPs for SAML based authentication:
+1) Okta
+2) Onelogin
+
+## SAML based session usage for the OKta IDP
+
+```python
+from avi.sdk.saml_avi_api import OktaSAMLApiSession
+# create Avi API Session
+api = OktaSAMLApiSession.get_session("10.10.10.42", "okta_username", "okta_password")
```
grastogi23: If you are using factory then it should be 
ApiSession.get_session("10.10.10.42", "okta_username", "okta_password", idp=OktaSAMLApiSession)

If you are not then it should simply be
OktaSAMLApiSession("10.10.10.42", "okta_username", "okta_password")

grastogi23: Please do the same for onelogin

python/avi/sdk/avi_api.py
```diff
@@ -442,6 +452,24 @@ def reset_session(self):
         self.headers = self.user_hdrs
         self.authenticate_session()
 
+    @staticmethod
+    def get_idp_class(idp):
```
grastogi23: why not just take the idp class as the input rather than converting it?

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,471 @@
...
+    """"
+        Extends the Request library's session object to provide helper
+        utilities to work with Avi Controller like SAML authentication,
+        api massaging
+        etc.
+    """
+
+    SAML_URL_SUFFIX = "/sso/login"
+
+    def __init__(self, controller=None, username=None, password=None,
```
grastogi23, line 310: it shouldn't need idp

python/avi/sdk/samples/virtualservice_examples_api.py
```diff
@@ -451,11 +452,20 @@ def clone_vs(self, vs_name, new_vip):
...
+    parser.add_argument('--idp', help="IDP type such as okta, onelogin, etc")
 
     args = parser.parse_args()
     print('parsed args', args)
     api = ApiSession.get_session(args.controller_ip, args.user, args.password,
-                 tenant=args.tenant, tenant_uuid=args.tenant_uuid)
+                                 tenant=args.tenant, tenant_uuid=args.tenant_uuid, idp=args.idp)
+
+    # Get controller session after successful SAML authentication from 'okta' IDP.
+    # api = OktaSAMLApiSession(args.controller_ip, args.user, args.password,
```
grastogi23: why are these commented out?

python/avi/sdk/avi_api.py
```diff
@@ -442,6 +452,24 @@ def reset_session(self):
...
+        if str(idp).lower() == "okta":
+            from avi.sdk.saml_avi_api import OktaSAMLApiSession
+            idp_class = OktaSAMLApiSession
+            logger.info("Using OktaSAMLApiSession to create "
+                        "controller session")
+        elif str(idp).lower() == "onelogin":
+            from avi.sdk.saml_avi_api import OneloginSAMLApiSession
+            idp_class = OneloginSAMLApiSession
```
grastogi23: I think we should let the caller pass the IDP class itself. This way it can be properly tested.

grastogi23 (changes requested):

python/avi/sdk/avi_api.py
gitshrikant: I thought it is more easy for the user to pass string idp value i.e. 'okta' or 'onelogin' and we can hide the complexity of instantiating corresponding IDP class. This can be done other way around as you commented. Modifying this according to the comments.

python/avi/sdk/samples/virtualservice_examples_api.py
gitshrikant: These are commented out just to show how SAML based controller session can be initiated. Otherwise, I can create a separate example for SAML based authentication.

gitshrikant: I will create seperate examl for SAML.

python/avi/sdk/avi_api.py
gitshrikant: Modified accordingly.

python/avi/sdk/saml_avi_api.py
gitshrikant: removed idp.

README.md
gitshrikant: Done.

README.md
```diff
@@ -84,7 +84,72 @@ user=os.environ.get('USER')
...
+for vs in resp.json()['results']:
+    print vs['name']
+
+# delete virtualservice
+resp = api.delete_by_name('virtualservice', 'sample_vs')
+```
+
+SAML session can also be invoked by following:
+```
+api = ApiSession.get_session("10.10.10.42", "onelogin_username", "onelogin_password", idp="OneloginSAMLApiSession")
```
grastogi23: change the example to pass the Class itself. OneloginSAMLApiSession

python/avi/sdk/avi_api.py
```diff
@@ -398,11 +402,14 @@ def get_session(
         :param retry_conxn_errors: retry on connection errors
```
grastogi23, line 403: Add documentation for idp. Ideally change idp to idp_class.

grastogi23: documentation should say the two IDP supported. may be check for validity of them here. The base class type should be ApiSession.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,469 @@
...
+                 token=None, tenant=None, tenant_uuid=None, verify=False,
+                 port=None, timeout=60, api_version=None,
+                 retry_conxn_errors=True, data_log=False,
+                 avi_credentials=None, session_id=None, csrftoken=None,
+                 lazy_authentication=False, max_api_retries=None,
+                 idp_cookies=None):
+        self.idp_cookies = idp_cookies
```
grastogi23, line 73: There are no doc strings for the arguments.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,469 @@
...
+            avi_credentials, session_id, csrftoken,
+            lazy_authentication, max_api_retries)
+        return
+
+    def saml_assertion(self, username, password):
+        """
+        Perform SAML request from controller to IDPs.
+        Establish session with controller and IDP.
+        Assert SAML request url into the response header.
+        :return: controller session and IDP response
```
grastogi23, line 92: add doc strings for username and password.

python/avi/sdk/saml_avi_api.py
```diff
@@ -0,0 +1,469 @@
...
+                    resp.status_code, resp.text), resp)
+        return controller_session, resp
+
+    def authenticate_session(self):
+        """
+        Performs session SAML authentication with Avi controller
+        and IDPs. Stores session cookies and sets header.
+        """
+
+        saml_response_regex = r'<input type=\"hidden\" ' \
```
grastogi23: These can be class variables. no need to create with every session.

grastogi23 (changes requested):

README.md
gitshrikant: changed!

PR #867, opened 2019-03-21
add more metrics to VS script

christiantreutler-avi:


python/avi/sdk/utils/waf_sizing/vs_metrics.py
```diff
@@ -0,0 +1,88 @@
+import argparse
```
mirkodziadzka-avi, line 2: To make it compatible with python3
```.py
from __future__ import print_function
```

python/avi/sdk/utils/waf_sizing/vs_metrics.py
```diff
@@ -0,0 +1,88 @@
...
+        'tenant': (args.tenant if args.tenant else 'admin'),
+        'step': args.step,
+        'limit': args.limit,
+        'entity_uuid': entity_uuid,
+        'pad_missing_data': False
+    }
+
+    rsp = api_utils.get_metrics_collection(tenant=args.tenant,
+                                           metric_requests=[mq])
+    print 'metrics query', mq
```
mirkodziadzka-avi: to make it compatible with python3, please use
```.py
print("metrics query {0}".format(mq))
```
or similar

python/avi/sdk/utils/waf_sizing/vs_metrics.py
```diff
@@ -0,0 +1,88 @@
...
+         "l7_client.sum_post_reqs"
+         )
+
+    mq = {
+        'metric_id': (args.metrics if args.metrics else metrics),
+        'tenant': (args.tenant if args.tenant else 'admin'),
+        'step': args.step,
+        'limit': args.limit,
+        'entity_uuid': entity_uuid,
+        'pad_missing_data': False
```
grastogi23: you can also pass include_name and include_ref:true. This will make the API return with name of VS as well. It might be useful

grastogi23 (approved):

python/avi/sdk/utils/waf_sizing/vs_metrics.py
chandrak6464: @christian. I think we we shouldnt print metrics query. We can comment out. We have the other script (parse_metrics.py) which parses this and the config file. If the output is not in json it will be a problem.

python/avi/sdk/utils/waf_sizing/vs_metrics.py
chandrak6464: Also, can we add the same thing to SE (se_metrics.py) as well. That way, we know the SE wise load and hence better estimation.

python/avi/sdk/utils/waf_sizing/vs_metrics.py
christiantreutler-avi: done

christiantreutler-avi:
@chandrak6464 I added the other files as well. Maybe I am not quite sure what the SE metrics are we are trying to collect. Somehow I thought we might collect the same as the VS, but these return 0, so they may not be available? What other SE metrics would be useful?

chandrak6464:
@christiantreutler-avi  SE metrics will give the same metrics per SE point of view. If VS's are disjointly scaled, it will be hard to size based on VS metrics alone. SE metrics should work.

christiantreutler-avi:
@chandrak6464 ah looks like I queried the SE, where the VS was not hosted and therefore had no data. The other SE had lots of data for me ;)

PR #877, opened 2019-04-08
Clone vs 1.1.1

thebertster:
Some tweaks were needed to handle sslkeyandcertificates better.


python/avi/sdk/samples/clone_vs.py
```diff
@@ -19,7 +19,7 @@
 
 DEFAULT_API_VERSION = '16.4.4'
 
-AVICLONE_VERSION = [1, 1, 0]
+AVICLONE_VERSION = [1, 1, 1]
```
grastogi23, line 22: @thebertster do you really need to version this separately.. the SDK is versioned as well.

grastogi23 (approved):

PR #916, opened 2019-05-13
AV-62611: Fixed the code to always send both sessionid and avi-sessionid. This makes the SDK compatible with both controller that is accepting avi-sessionid as well older ones that only support sessionid

grastogi23:


chaitanyaavi (approved):

grastogi23 (commented):
merging it after updated cassettes

PR #920, opened 2019-05-17
Handle pacthed names during vs_filtering

sandeepavi:
The PR will handle the case where patch and config migration are expected to be done in single command

Tested the below three cases and they are working as expected.

- Only vs_filter option is given.
Ex: `./f5_converter.py -f ~/avi/a.conf --vs_filter fmr-jenkins-m1-np2107.fmrco.com-4444-mmk,fmr-jenkins-m1-np2107.fmrco.com-80-mmk -t PR000001 -o /tmp  --controller_version 17.2.14`
- Only patch Option is given.
Ex:  `./f5_converter.py -f ~/avi/a.conf -t PR000001 -o /tmp --patch ~/avi/patch1.yaml --controller_version 17.2.14`
- Both vs_filter and patch option are give.
`./f5_converter.py -f ~/avi/a.conf --vs_filter fmr-jenkins-m1-np2107.fmrco.com-4444-mmk,fmr-jenkins-m1-np2107.fmrco.com-80-mmk -t PR000001 -o /tmp --patch ~/avi/patch1.yaml --controller_version 17.2.14`

@grastogi23 @chaitanyaavi Please review the PR and let me know if there are any changes required.


python/avi/migrationtools/avi_converter.py
```diff
@@ -41,8 +41,26 @@ def process_for_utils(self, avi_config):
...
+
         if self.vs_filter:
-            avi_config = filter_for_vs(avi_config, self.vs_filter)
+            # Check if patch data is provided and if any vs is getting patched
+            # Create a map of all vs names and their respective patched names from patch data
+            vs_names = self.vs_filter
+            if self.patch and patches.get('VirtualService'):
+                vs_patched_names = {}
+                for vs in patches['VirtualService']:
+                    vs_patched_names[vs['match_name']] = vs['patch'].get('name')
```
grastogi23: we should also do it for other options like match_name_regex and match_name_list.

sandeepavi: Thanks @grastogi23 for pointing them out. Will handle them as well.

chaitanyaavi: If user is writing the patch file isn't it expected that user should use the patched new name in the filter list? We do patching first and then run filter script.

sandeepavi: Thanks @chaitanyaavi. Seems like that is the case. I guess we just have to update the documentation with the same details so that there is no confusion in usage. I will close this PR.

PR #931, opened 2019-06-04
Assets updated for go models and clients

patilshrikant786:


grastogi23 (approved):

grastogi23:
we don't need to update the models as this was created out of 18.2.3 tree.

PR #932, opened 2019-06-04
Assets updated for swagger specs

patilshrikant786:


grastogi23:
no need to udpate these.

PR #934, opened 2019-06-06
AV-63140 Added tenant switching functionality in avisession.

gitshrikant:
Updated create VS example.

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -171,10 +171,14 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 }
 
 //SwitchTenant Sets tenant into the avisession.
-func (avisess *AviSession) SwitchTenant(tenant string) (error) {
-	avisess.tenant = tenant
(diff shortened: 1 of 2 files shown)


grastogi23:
we will also need functionality to change it on every call. let us discuss it in the review.

grastogi23 (changes requested):
this needs further discussion

go/session/avisession.go
```diff
@@ -170,6 +170,15 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 	return avisess, err
 }
 
+//SwitchTenant Sets tenant into the avisession.
+func (avisess *AviSession) SwitchTenant(tenant string) (error) {
+	avisess.tenant = tenant
+	if avisess.tenant == "" {
```
grastogi23: we should not overwrite the avisess.tenant. intead if tenant is nil then simply return error.

grastogi23 (changes requested):

PR #955, opened 2019-07-24
AV-65723 #comment Add method to do get_all operaion

sabandi:


python/avi/sdk/avi_api.py
```diff
@@ -931,6 +931,35 @@ def get_obj_uuid(self, obj):
...
+
+        page = 2 #  Initialized to 2 for new page requests
+        while resp.get('next', None) is not None:
+            params['page'] = page
+            resp = self.get(obj_type, tenant, tenant_uuid, timeout=timeout,
+                            params=params, api_version=api_version, **kwargs).json()
+            all_objs += resp['results']
+            page += 1
+
+        for obj in all_objs:
```
sabandi: @grastogi23  I am not sure what benefit we get by using this generator model as we anyways have all the objects in memory.

We could do a simple `return all_objs` .
Please let me know your thoughts.

grastogi23: we should not keep all objects in the memory... The semantics is to provide an iterator not fetch all objects.

python/avi/sdk/avi_api.py
```diff
@@ -931,6 +931,35 @@ def get_obj_uuid(self, obj):
         else:
             return None
 
+    def get_all_objects_of_type(self, obj_type, tenant='', tenant_uuid='',
+                                timeout=None, params=None, api_version=None,
```
grastogi23: should document the parameters. Also provide an example use of this api.

let us call it get_objects_iter so it is explicit that it is not return all objects but an iterator to iterate over objects.

grastogi23 (changes requested):
please see inline comments.

python/avi/sdk/avi_api.py
```diff
@@ -931,6 +931,35 @@ def get_obj_uuid(self, obj):
...
+                                **kwargs):
+        """
+        Helper method to get all objects of specific type from controller.
+        """
+        all_objs = []
+        if params is None:
+            params={}
+        params['page_size'] = 200 # Max limt on Controller
```
grastogi23: let us take the page_size from the params as well. But, we can check for max of 200

python/avi/sdk/avi_api.py
```diff
@@ -931,6 +931,35 @@ def get_obj_uuid(self, obj):
...
+            all_objs += resp['results']
+        else:
+            # For apis returning single object eg. api/cluster
+            all_objs += [resp]
+
+        page = 2 #  Initialized to 2 for new page requests
+        while resp.get('next', None) is not None:
+            params['page'] = page
+            resp = self.get(obj_type, tenant, tenant_uuid, timeout=timeout,
+                            params=params, api_version=api_version, **kwargs).json()
```
grastogi23: what happens in case of error. It should raise exception?

python/avi/sdk/avi_api.py
```diff
@@ -931,6 +931,83 @@ def get_obj_uuid(self, obj):
...
+        on top of the pagination functionality provided by Avi Controller.
+
+        :param *args: Accepts all args accepted by get method of class
+            ApiSession.
+        :params **kwargs: Accepts all kwargs accepted by get method of class
+            ApiSession.
+
+        Yields batch of objects in the range of 0 to page_size passed in kwargs.
+        """
+        resp = self.get(*args, **kwargs).json()
```
sabandi, line 946: .json() method in avi_api.py checks and raises various exceptions. Hence left the exception handling as it is already taken care of in json() method @ https://github.com/avinetworks/sdk/blob/bba674f45ce80374f2491a24b2137050cf86f70a/python/avi/sdk/avi_api.py#L89

grastogi23 (approved):

chaitanyaavi (approved):

grastogi23 (approved):

PR #970, opened 2019-09-05
AV-62249 F5 migration tool imported invalid server listening ports

Rohan-sss1:


python/avi/migrationtools/f5_converter/pool_converter.py
```diff
@@ -168,6 +168,10 @@ def create_pool_object(self, name, desc, servers, pd_action, algo,
             'lb_algorithm': algo,
             'cloud_ref': conv_utils.get_object_ref(cloud_ref, 'cloud')
         }
+        for server in servers:
+            if server["port"] == '0':
+                pool_obj.update({"use_service_port": "true"})
+                server.update(port = "80")
```
grastogi23: let us just remove the server.update line. also, the loop should be broken. You can refactor the code to use "any" builtin

grastogi23 (approved):

PR #971, opened 2019-09-06
AV-43034 WAF vpatch script

JacekKrzeminski:
Python script to load Qualys Web and ZAP scan results into AVI controller.
It creates a dedicated PSM group for each supported scanner type, and a PSM rule for each discovered vulnerability of supported type.

### DIFF
--- a/python/avi/sdk/utils/dast/avi-iwaf-vpatch.py
+++ b/python/avi/sdk/utils/dast/avi-iwaf-vpatch.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+
+import sys
+if sys.version_info < (3, 5):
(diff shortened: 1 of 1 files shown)


JacekKrzeminski:
Please advice if:
- this is the right place to merge the script
- Pipfile should go here as well
- README should go here, in addition to a KB article

JacekKrzeminski:
https://avinetworks.atlassian.net/wiki/spaces/EN/pages/827719977/DAST+scanner+integrations+-+implementation

https://avinetworks.atlassian.net/wiki/spaces/EN/pages/555221081/DAST%2Bscanner%2Bintegrations%2B-%2Bvirtual%2Bpatching

grastogi23 (approved):

PR #984, opened 2019-09-25
Travis 18.2.6

chaitanyaavi:


python/version.py
```diff
@@ -4,7 +4,7 @@
 '''
 
 AVI_VERSION = '18.2.6'
-AVI_PIP_VERSION = '18.2.6b2'
+AVI_PIP_VERSION = '18.2.6'
```
grastogi23: this shouldn't be changed.

grastogi23 (changes requested):

grastogi23 (approved):

PR #1025, opened 2019-11-13
AV-66090 Added fixes to handle systemconfiguration dns vs ref

yograjshisode:
Also added fixes to handle deletion of the default objects, sytemconfiguration, cloudproperties, controllrproperties, seproperties.

### DIFF
--- a/python/avi/sdk/utils/ansible_utils.py
+++ b/python/avi/sdk/utils/ansible_utils.py
@@ -45,7 +45,8 @@ def ansible_return(module, rsp, changed, req=None, existing_obj=None,
     Returns: specific ansible module exit function
     """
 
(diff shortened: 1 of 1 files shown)


python/avi/sdk/utils/ansible_utils.py
```diff
@@ -467,8 +470,9 @@ def avi_ansible_api(module, obj_type, sensitive_fields):
         if rsp:
             if rsp.status_code == 204:
                 changed = True
-            else:
+            elif not any(error in str(rsp.text) for error in SKIP_DELETE_ERROR):
```
grastogi23, line 473: check with @gcox1

grastogi23 (approved):

PR #1059, opened 2019-12-20
AV-71800 avi iwaf vpatch tests master

JacekKrzeminski:
Revive the version of the script from fb0ba79298bb10b569f3dedddd38bbc1420d74ff.
Add a unit test. Fix an issue found by the test.

### DIFF
--- a/python/avi/migrationtools/common/dast/test_vpatch.py
+++ b/python/avi/migrationtools/common/dast/test_vpatch.py
@@ -1,0 +1,651 @@
+import xmltodict
+dast=__import__("avi-iwaf-vpatch")
+
+
+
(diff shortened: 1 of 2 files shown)


python/avi/migrationtools/common/dast/test_vpatch.py
```diff
@@ -0,0 +1,651 @@
+import xmltodict
+dast=__import__("avi-iwaf-vpatch")
```
JacekKrzeminski, line 2: It's embarrassing but we gave the script a name which is not valid for a python module. (I am still learning it).

grastogi23 (approved):

PR #1063, opened 2020-01-03
AV-74124 Added changes to avoid unnecessary patch(delete) call

yograjshisode:


grastogi23:
would this need to be pushed to the core ansible as well? @chaitanyaavi

grastogi23 (approved):

PR #1081, opened 2020-01-29
AV-76352 Fixed avi_obj_cmp to allow and detect change for empty list if existing object have something in that list

yograjshisode:


grastogi23:
we need to commit to ansible core as well. and we need to add it to the unit test.

grastogi23 (changes requested):
please add unit test.

grastogi23 (approved):

PR #1103, opened 2020-03-03
AV-70462 Added change to wait for the controller ready state

yograjshisode:


go/session/avisession.go
```diff
@@ -436,6 +436,17 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 
 	resp, err := avisess.client.Do(req)
 	if err != nil {
+		// Wait untill controller is in ready state
+		if strings.Contains(err.Error(), "connection refused") == true {
+			check, err := avisess.CheckControllerStatus()
```
grastogi23: should refactor to better Go standards.

grastogi23 (changes requested):
please add a unit test for the timeout.

go/session/avisession.go
grastogi23: if strings.Contains(err.Error(), "connection refused") {
	if check, err := avisess.CheckControllerStatus(); err == nil && check {
        	if uri == "login" {
			avisess.initiateSession()
		} else {
			return avisess.restRequest(verb, uri, payload, tenant, errorResult, retry+1)
		}
       } else {
          return nil, err;
       }

grastogi23 (changes requested):

go/examples/test/aviclient_test.go
```diff
@@ -0,0 +1,27 @@
...
+		session.SetInsecure, session.SetLazyAuthentication(true))
+	if err != nil {
+		fmt.Println("Issue with lazyauthentication")
+		t.Fail()
+	}
+	hmData, err := aviClient.HealthMonitor.GetAll()
+	if hmData == nil || err != nil {
+		fmt.Println("Got expected error for invalid controller IP. [ERROR]:", err.Error())
+	} else {
+		fmt.Println("Didn't got the error for invalid controller IP.")
```
grastogi23: change got to "get"

go/session/avisession.go
```diff
@@ -436,6 +436,18 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 
 	resp, err := avisess.client.Do(req)
 	if err != nil {
+		// Wait untill controller is in ready state
+		if strings.Contains(err.Error(), "connection refused") {
+			if check, err := avisess.CheckControllerStatus(); err == nil && check {
+				if uri == "login" {
+					avisess.initiateSession()
```
grastogi23: This is not correct. should this be return? I don't think tests are exercising this part of the code.

grastogi23 (changes requested):

go/session/avisession.go
yograjshisode: Added fix for the return. Also verified the test case and found that controller with invalid IP not hitting the this part of code. We have also added the test case for terraform playbook in avi-dev which will help us to test this part of code. It's hard to reproduce the scenario for connection refuse once the controller is in up state. I have tried different ways like reboot warm and clean using API, disabled the avi_portal service, reboot VM(having issue with IP allocation)

go/examples/test/aviclient_test.go
```diff
@@ -0,0 +1,27 @@
...
+		fmt.Println("Issue with lazyauthentication")
+		t.Fail()
+	}
+	hmData, err := aviClient.HealthMonitor.GetAll()
+	if hmData == nil || err != nil {
+		fmt.Println("Got expected error for invalid controller IP. [ERROR]:", err.Error())
+	} else {
+		fmt.Println("Didn't get the error for invalid controller IP.")
+		t.Fail()
+	}
```
grastogi23, line 26: we should also add test for the case that initiate session works. 

1. create session with a new password that is not yet set on controller. This should fail
2. now update the password using another session to the new password
3. try the API with the correct password again. there should be no error.

grastogi23 (changes requested):

go/session/avisession.go
```diff
@@ -436,6 +436,19 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 
 	resp, err := avisess.client.Do(req)
 	if err != nil {
+		// Wait untill controller is in ready state
```
grastogi23, line 440: if err = nil {
glog.Infof("Req for uri %v RespCode %v", url, resp.StatusCode)
	errorResult.HttpStatusCode = resp.StatusCode
	avisess.collectCookiesFromResp(resp)
} else if ("connection refused") {
reetry
} else {
errorResult.err = fmt.Errorf("client.Do uri %v failed: %v", uri, err)
		dump, err := httputil.DumpRequestOut(req, true)
		debug(dump, err)
		return result, errorResult
}

grastogi23 (approved):

PR #1128, opened 2020-03-31
AV-81062: AviCollectionResult add Next field for next url

chauhanshubham:


grastogi23 (changes requested):
looks good. let us add test case for it and need to add code to populate it too.

go/session/avisession.go
```diff
@@ -755,6 +755,7 @@ func convertAviResponseToMapInterface(resbytes []byte) (interface{}, error) {
 type AviCollectionResult struct {
 	Count   int
 	Results json.RawMessage
+	Next	string
```
grastogi23, line 758: I see this being added but I don't see it being set.

grastogi23: is this automatic via unmarshall?

grastogi23 (approved):

PR #1131, opened 2020-04-06
AV-70769 Added changes to return error from refreshAuthToken method

yograjshisode:


go/session/avisession.go
```diff
@@ -274,13 +277,13 @@ func (avisess *AviSession) setAuthToken(authToken string) error {
 }
 
 // SetAuthToken - Use this for NewAviSession option argument for setting authToken
-func SetRefreshAuthTokenCallback(f func() string) func(*AviSession) error {
+func SetRefreshAuthTokenCallback(f func() (string, error)) func(*AviSession) error {
```
grastogi23: if avisess.refreshAuthTokenCallback != nil {
			if token, err:= refreshAuthTokenCallback(); err != nil {
				avisess.setAuthToken(avisess.refreshAuthToken())
			} else {
				return err
			}
		}

grastogi23 (changes requested):
Need to add another callback function that returns error as well.

grastogi23 (approved):

PR #1141, opened 2020-04-28
expose RestRequest to return http.Response

chauhanshubham:
The exposed RestRequest provides a way to return the whole http response (including headers), and not just the response body. This is mostly required by avi_api_proxy and other clients that would need to read headers returned by AVI api server. Currently there seems to be no way to fetch the http response in the go SDK, whereas the python SDK provides the same functionality.
This function does not change the existing restRequest implementation.
cc: @grastogi23

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"github.com/golang/glog"
 	"io"
 	"io/ioutil"
 	"math"
@@ -18,6 +17,8 @@ import (
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/golang/glog"
 )
 
 type AviResult struct {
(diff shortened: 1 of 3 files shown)


go/session/avisession.go
```diff
@@ -457,6 +458,85 @@ func (avisess *AviSession) collectCookiesFromResp(resp *http.Response) {
 	}
 }
 
+// ProxyRestRequest makes a REST request to the Avi Controller's REST API.
+// Returns http.Response if successful
+// Note: The caller of the function is responsible for doing resp.Body.Close()
+func (avisess *AviSession) ProxyRestRequest(verb string, uri string, payload interface{}, tenant string, lastError error, retryNum ...int) (*http.Response, error) {
```
grastogi23: There are lot of code duplication. Can you please explore if we can merge the two. Also, do we need to create a multi-part variant as well? I can't remember why we needed this. can you please add it to the descriiptioin of the functiion

grastogi23 (changes requested):

go/session/avisession.go
chauhanshubham: updated the PR comment

grastogi23 (approved):

avi-ashutosh:
Looks good to me, thanks for the enhancement @shubhamavi

grastogi23 (approved):

PR #1162, opened 2020-05-14
Rc clone vs dev

thebertster:
Some additional functionality and fixes and have added test to test suite as previously discussed. I am told that we no longer need to mark the test with a TCID within the test suite and this is done externally, so I do not have a @pytest.mark.TCIDxxxxxx decorator for my test function. I trust this is correct.


grastogi23 (approved):
let us wait for @chaitanyaavi  to review as well.

grastogi23 (approved):

PR #1169, opened 2020-06-05
AV-86846  handle payload in post

manojkumarjain:
AV-86846  handle payload in post

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -950,6 +950,8 @@ func (avisess *AviSession) GetRaw(uri string) ([]byte, error) {
 
 // PostRaw performs a POST API call and returns raw data
 func (avisess *AviSession) PostRaw(uri string, payload interface{}) ([]byte, error) {
+        glog.Infof("MANOJ PostRaw: %v", payload)
+        resp, rerror := avisess.restRequest("POST", uri, payload, "", nil)
 	resp, rerror := avisess.restRequest("POST", uri, nil, "", nil)
 	if rerror != nil || resp == nil {
 		return nil, rerror
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -950,12 +950,10 @@ func (avisess *AviSession) GetRaw(uri string) ([]byte, error) {
 
 // PostRaw performs a POST API call and returns raw data
 func (avisess *AviSession) PostRaw(uri string, payload interface{}) ([]byte, error) {
-        glog.Infof("PostRaw: %v", payload)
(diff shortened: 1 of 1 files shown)


go/session/avisession.go
```diff
@@ -950,6 +950,8 @@ func (avisess *AviSession) GetRaw(uri string) ([]byte, error) {
 
 // PostRaw performs a POST API call and returns raw data
 func (avisess *AviSession) PostRaw(uri string, payload interface{}) ([]byte, error) {
+        glog.Infof("MANOJ PostRaw: %v", payload)
```
grastogi23: please remove "Manoj" from the debug. Also, printing payload may be very expensive. Let us also add unit test for this. I am surprised this was not caught in the previous commit.

grastogi23 (changes requested):

go/session/avisession.go
manojkumarjain: Will do after testing on ctrl

grastogi23: let us may be create a jira for sandeep or sdk to add support for PutRaw as well. I checked that is not supported.

go/session/avisession.go
```diff
@@ -950,6 +950,8 @@ func (avisess *AviSession) GetRaw(uri string) ([]byte, error) {
...
+        resp, rerror := avisess.restRequest("POST", uri, payload, "", nil)
 	resp, rerror := avisess.restRequest("POST", uri, nil, "", nil)
```
chauhanshubham, line 955: I think replacing nil with payload here is the only thing required. Newline at L954 needs to be replaced at L955

go/session/avisession.go
```diff
@@ -950,7 +950,8 @@ func (avisess *AviSession) GetRaw(uri string) ([]byte, error) {
 
 // PostRaw performs a POST API call and returns raw data
 func (avisess *AviSession) PostRaw(uri string, payload interface{}) ([]byte, error) {
-	resp, rerror := avisess.restRequest("POST", uri, nil, "", nil)
+        glog.Infof("PostRaw: %v", payload)
```
chauhanshubham: not sure if we should log this, considering we are not logging the Post() payload as well. Maybe change it to Debug or remove it for consistency across other functions

go/session/avisession.go
manojkumarjain: Sure Gaurav

manojkumarjain:
@grastogi23  please approve and merge.
Unit testing completed

chauhanshubham (approved):

grastogi23 (approved):

PR #1176, opened 2020-06-16
Amend clone_vs test for Jenkins

thebertster:


thebertster:
Making this change as part of another PR.

PR #1184, opened 2020-06-30
AV-88850: Provide more GO SDK client controls

mayank-avinetworks:
**Current Behaviour of SDK to poll controller status:**
     Total 10 iterations, where each for iteration the sleep time increases exponentially with formula ```math.Exp(float64(num_iteration))*3``` . 
**New Behaviour**:
     During client init, the number of retries and the time interval between each retry can be set.
    ```aviClient, err := clients.NewAviClient("10.10.25.25", "admin",
		session.SetPassword("something"),
		session.SetTenant("admin"),
                SetControllerStatusCheckLimits(5, 10), // numRetries, Time interval in each retry.
		session.SetInsecure)```
     If SetControllerStatusCheckLimits(xx,yy) is not used, the default behaviour of retrying for 10 times over the span of 5 mins will be invoked.

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -127,9 +127,9 @@ type AviSession struct {
 	// optional api retry interval in milliseconds
 	api_retry_interval int
 
-	// Number of retries the SDK should attempt when controller is not reachable.
+	// Number of retries the SDK should attempt to check controller status.
 	ctrlStatusCheckRetryCount int
-	// Total number of seconds to wait before attemptemptin another try to reach to controller.
+	// Time interval in seconds within each retry to check controller status.
 	ctrlStatusCheckRetryInterval int
 }
 
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -130,7 +130,12 @@ type AviSession struct {
 
 	// flag to check if on API failure, the controller status should be polled by
 	// sleeping for exponentially increasing time per interation or not.
-	allowInfiniteCtrlStatusCheck bool
+	infinitelyCheckCtrlStatus bool
+
+	// Number of retries the SDK should attempt when controller is not reachable.
+	ctrlStatusCheckRetryCount int
+	// Total number of seconds to wait before attemptemptin another try to reach to controller.
+	ctrlStatusCheckRetryInterval int
 }
 
 const DEFAULT_AVI_VERSION = "17.1.2"
@@ -138,7 +143,6 @@ const DEFAULT_API_TIMEOUT = time.Duration(60 * time.Second)
 const DEFAULT_API_TENANT = "admin"
 const DEFAULT_MAX_API_RETRIES = 3
 const DEFAULT_API_RETRY_INTERVAL = 500
-const DEFAULT_NON_EXPON_STAT_INTERVAL = time.Duration(30 * time.Second)
 
 //NewAviSession initiates a session to AviController and returns it
 func NewAviSession(host string, username string, options ...func(*AviSession) error) (*AviSession, error) {
@@ -155,9 +159,10 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 	avisess.tenant = ""
 	avisess.insecure = false
 	// This was the default behavior before providing the flexibility to configure session.
-	// This can be over-ridden while client init.
-	avisess.allowInfiniteCtrlStatusCheck = true
-
+	avisess.infinitelyCheckCtrlStatus = true
+	// The default behaviour was for 10 iterations, if client does not init session with specific retry
+	// count option the controller status will be checked 10 times.
+	avisess.ctrlStatusCheckRetryCount = 10
 	for _, option := range options {
 		err := option(avisess)
 		if err != nil {
@@ -335,11 +340,15 @@ func SetInsecure(avisess *AviSession) error {
 	return nil
 }
 
-// SetQuickControllerStatusCheck informs the SDK to not to poll for the controller status
-// by sleeping exponentially on every polling iteration during API failure.
-func SetQuickControllerStatusCheck(avisess *AviSession) error {
-	avisess.allowInfiniteCtrlStatusCheck = false
-	return nil
+// SetControllerStatusCheckLimits allows client to limit the number of tries the SDK should
+// attempt to reach the controller at the time gap of specified time intervals.
+func SetControllerStatusCheckLimits(numRetries, numIntervals int) func(*AviSession) error {
+	return func(avisess *AviSession) error {
+		avisess.infinitelyCheckCtrlStatus = false
+		avisess.ctrlStatusCheckRetryCount = numRetries
+		avisess.ctrlStatusCheckRetryInterval = numIntervals
+		return nil
+	}
 }
 
 // SetTransport - Use this for NewAviSession option argument for configuring http transport to enable connection
@@ -548,7 +557,7 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 		}
 	}
 	if retryReq {
-		check, err := avisess.CheckControllerStatus(avisess.allowInfiniteCtrlStatusCheck)
+		check, err := avisess.CheckControllerStatus()
 		if check == false {
 			glog.Errorf("restRequest Error during checking controller state %v", err)
 			return nil, err
@@ -683,7 +692,7 @@ func (avisess *AviSession) restMultipartUploadRequest(verb string, uri string, f
 	}
 
 	if retryReq {
-		check, err := avisess.CheckControllerStatus(avisess.allowInfiniteCtrlStatusCheck)
+		check, err := avisess.CheckControllerStatus()
 		if check == false {
 			glog.Errorf("restMultipartUploadRequest Error during checking controller state")
 			return err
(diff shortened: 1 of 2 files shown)


go/session/avisession.go
```diff
@@ -327,6 +340,17 @@ func SetInsecure(avisess *AviSession) error {
 	return nil
 }
 
+// SetControllerStatusCheckLimits allows client to limit the number of tries the SDK should
+// attempt to reach the controller at the time gap of specified time intervals.
+func SetControllerStatusCheckLimits(numRetries, numIntervals int) func(*AviSession) error {
+	return func(avisess *AviSession) error {
```
mayank-avinetworks, line 340: Shall we sanitise the input from user against +ve integer check?

go/session/avisession.go
```diff
@@ -327,6 +340,17 @@ func SetInsecure(avisess *AviSession) error {
 	return nil
 }
 
+// SetControllerStatusCheckLimits allows client to limit the number of tries the SDK should
+// attempt to reach the controller at the time gap of specified time intervals.
+func SetControllerStatusCheckLimits(numRetries, numIntervals int) func(*AviSession) error {
```
yograjshisode: Can we add a argument to set the value of  infinitelyCheckCtrlStatus in function SetControllerStatusCheckLimits? It will help to easily switch between the infinitelyCheckCtrlStatus mode and newer one using the same session

mayank-avinetworks: @yograjshisode : I thought of few situations if we make that as configurable.
1. To maintain the backward compatibility we keep the same number of iterations and the sleep behaviour as it was earlier.
2. If we allow to change the mode in between the same session, then someone can change the configured retry/interval counts to sleep for exponentially increasing time on controller status check poll, which changes our behaviour the way it polls by sleeping exponentially longer time, today.

manojkumarjain: Looks fine to me.
Should we ask Sudipto  to pull SDK from this branch and test it

go/session/avisession.go
```diff
@@ -127,6 +127,15 @@ type AviSession struct {
 
 	// optional api retry interval in milliseconds
 	api_retry_interval int
+
+	// flag to check if on API failure, the controller status should be polled by
+	// sleeping for exponentially increasing time per interation or not.
+	infinitelyCheckCtrlStatus bool
```
manojkumarjain: Please change variable name

go/session/avisession.go
```diff
@@ -127,6 +127,15 @@ type AviSession struct {
...
+
+	// Number of retries the SDK should attempt when controller is not reachable.
```
manojkumarjain: The comments seems to be incorrect, CRL  should be reachable 
Also change var name 
ctrlStatusCheckRetryCount --> retryCount 
ctrlStatusCheckRetryInterval --> retryInterval

go/session/avisession.go
```diff
@@ -149,7 +158,11 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 	avisess.prefix = "https://" + avisess.host + "/"
 	avisess.tenant = ""
 	avisess.insecure = false
-
+	// This was the default behavior before providing the flexibility to configure session.
+	avisess.infinitelyCheckCtrlStatus = true
```
manojkumarjain: infinitelyCheckCtrlStatus --> isReachable or something similar

go/session/avisession.go
manojkumarjain: I agree, let us keep it backward compatible, if options are not set

go/session/avisession.go
manojkumarjain: Yes all user supplied should be sanitize

go/session/avisession.go
```diff
@@ -327,6 +340,17 @@ func SetInsecure(avisess *AviSession) error {
 	return nil
 }
 
+// SetControllerStatusCheckLimits allows client to limit the number of tries the SDK should
+// attempt to reach the controller at the time gap of specified time intervals.
+func SetControllerStatusCheckLimits(numRetries, numIntervals int) func(*AviSession) error {
+	return func(avisess *AviSession) error {
+		avisess.infinitelyCheckCtrlStatus = false
```
manojkumarjain: Do we need this infinitelyCheckCtrlStatus?

go/session/avisession.go
```diff
@@ -805,12 +829,11 @@ func debug(data []byte, err error) {
...
 
 //Checking for controller up state.
-//This is an infinite loop till the controller is in up state.
-//Return true when controller is in up state.
+// Flexible to wait on controller status infinitely or for fixed time span.
 func (avisess *AviSession) CheckControllerStatus() (bool, error) {
 	url := avisess.prefix + "/api/cluster/status"
-	//This is an infinite loop. Generating http request for a login URI till controller is in up state.
-	for round := 0; round < 10; round++ {
+	var isControllerUp bool
```
manojkumarjain, line 831: Initialize to false

go/session/avisession.go
grastogi23: We don't need this flag. if ctrlStatusCheckRetryInterval is not set then it is exponential. else it is fixed status check retry interval.

go/session/avisession.go
```diff
@@ -127,6 +127,15 @@ type AviSession struct {
 
 	// optional api retry interval in milliseconds
 	api_retry_interval int
+
+	// flag to check if on API failure, the controller status should be polled by
+	// sleeping for exponentially increasing time per interation or not.
+	infinitelyCheckCtrlStatus bool
+
+	// Number of retries the SDK should attempt when controller is not reachable.
+	ctrlStatusCheckRetryCount int
```
grastogi23, line 132: we already have retry count "max_api_retries". let us re-use it. no need to introduce a new retry number.

go/session/avisession.go
```diff
@@ -127,6 +127,15 @@ type AviSession struct {
...
+	// Total number of seconds to wait before attemptemptin another try to reach to controller.
+	ctrlStatusCheckRetryInterval int
```
grastogi23, line 134: Even this one can't we simply use api_retry_interval flag? In fact rename it to the better GoStandards.

go/session/avisession.go
```diff
@@ -830,10 +854,15 @@ func (avisess *AviSession) CheckControllerStatus() (bool, error) {
 			glog.Errorf("CheckControllerStatus Error while generating http request %v %v", url, err)
 		}
 		//wait before retry
-		time.Sleep(time.Duration(math.Exp(float64(round))*3) * time.Second)
+		if avisess.infinitelyCheckCtrlStatus {
```
grastogi23: we can simply make it adaptive that if retry interval is set it becomes static. If not then it is exponential. 

Also, I would like to cap it at something like 5 mins. There is no point in letting it grow for anything more than 5mins.

grastogi23 (changes requested):

go/session/avisession.go
mayank-avinetworks: sure, we can get rid of this.

go/session/avisession.go
mayank-avinetworks: @grastogi23 : I see few issues using the existing variables `api_retry_interval` and `max_api_retries` for controller status check polling.
The existing variables are used for timeout and retry count of HTTP client which actually sends request to controller.
* Controller bring up/reboot/re-electoion can take up to couple of mins, which controller status poll request might need to poll for significantly greater time period than a normal HTTP request.
* If  a user configures these variables during session init by considering that this will poll controller status for around 10 times over the period of  5/10 mins then it will also retry the failed REST APIs to controller for same time intervals and iterations.
* This could be other way as well. If use configures  3 retries over 10 sec each. that might not be able to poll controller status properly.

please let me know what are you thoughts on this?

go/session/avisession.go
mayank-avinetworks: sure, Manoj.

go/session/avisession.go
mayank-avinetworks: This was to override default value, but anyways, this flag is going away now :)

go/session/avisession.go
mayank-avinetworks: this will be initialised to false by itself .

manojkumarjain: isControllerUp := bool

go/session/avisession.go
manojkumarjain: Good point, let cap it to max 5 mins in case of exponential

go/session/avisession.go
mayank-avinetworks: go runtime initialises the bool to false by default.

manojkumarjain (approved):
Approve

go/session/avisession.go
grastogi23: Sure. makes sense... let us keep them separate.

go/session/avisession.go
```diff
@@ -829,11 +848,19 @@ func (avisess *AviSession) CheckControllerStatus() (bool, error) {
...
 			glog.Errorf("CheckControllerStatus Error while generating http request %v %v", url, err)
 		}
-		//wait before retry
-		time.Sleep(time.Duration(math.Exp(float64(round))*3) * time.Second)
+		// if controller status check interval is not set during client init, use the default SDK
+		// behaviour.
+		if avisess.ctrlStatusCheckRetryInterval == 0 {
+			// Default controller status check behaviour is 10 iterations at interval of 30 secs each.
+			// i.e total 5 mins.
+			time.Sleep(time.Duration(30) * time.Second)
```
grastogi23: min (time.Sleep(time.Duration(math.Exp(float64(round))*3) * time.Second), time.Sleep(time.Duration(30) * time.Second))

grastogi23 (changes requested):

go/session/avisession.go
mayank-avinetworks: Done. 👍

grastogi23 (approved):

PR #1219, opened 2020-08-04
Handled f_internal and f_hidden fields in sdk models

shardullatkar:


java/src/com/vmware/avi/sdk/model/Server.java
```diff
@@ -50,6 +50,9 @@
     @JsonProperty("ip")
     private IpAddr ip = null;
 
+    @JsonProperty("is_static")
```
chaitanyaavi: This should be static not is_static

shardullatkar: Resolved.

chaitanyaavi (approved):

PR #1235, opened 2020-08-18
Auto updated go models and clients from eng

shounak-opcito:


yograjshisode:
changes merged using PR: https://github.com/avinetworks/sdk/pull/1238

PR #1240, opened 2020-08-27
clone_vs.py: Add support for cloning GSLB Services

thebertster:


python/avi/sdk/samples/clone_vs.py
```diff
@@ -2568,6 +2764,65 @@ def clone_vs(self, old_vs_name, new_vs_name, enable_vs=False,
                     if args.tocloud:
                         print('%10s: %s' % ('Cloud', args.tocloud))
                     print()
+            elif args.obj_type == 'gs':
```
grastogi23, line 2778: can we split this into its own method/function so it gets own importance ;)

python/avi/sdk/samples/clone_vs.py
```diff
@@ -2568,6 +2764,65 @@ def clone_vs(self, old_vs_name, new_vs_name, enable_vs=False,
...
+                    # GS rather than values per new GS
+                    fqdns = [fqdns]
+                else:
+                    # Otherwise, make sure we have the same number of FQDNs,
+                    #  as the number of provided VS names
+
+                    if (len(fqdns) == num_new_gs):
+                        fqdns = [[fqdn] for fqdn in fqdns]
+                    else:
+                        raise Exception('Number of FQDNs '
```
grastogi23: good to print out the fqdns and the value that were not consistent.

python/avi/sdk/samples/clone_vs.py
```diff
@@ -2568,6 +2764,65 @@ def clone_vs(self, old_vs_name, new_vs_name, enable_vs=False,
...
+                               ' ['+args.totenant+']'
+                               if args.totenant else ''),
+                            flush=True)
+                    new_gs, cloned_objs, warnings = cl.clone_gs(
+                             old_gs_name=args.gs_name, new_gs_name=new_gs_name,
+                             enable_gs=args.enable, new_fqdns=new_fqdns,
+                             tenant=args.tenant, other_tenant=args.totenant,
+                             force_clone=force_clone)
+                    all_created_objs.append(new_gs)
+                    all_created_objs.extend(cloned_objs)
```
grastogi23, line 2820: what happens to the GSLB Pool members? those don't seem to be cloned? would that not cause problem if done on the same controller?

grastogi23 (changes requested):

python/avi/sdk/samples/clone_vs.py
thebertster: GSLB cloning is in its own method already (clone_gs).

python/avi/sdk/samples/clone_vs.py
thebertster: Well unlike a normal VS that has pools as referred objects, a GSLB VS just holds references to site and VS UUIDs within itself, so there are no child objects to clone.

It would be nice to have an option to recursively clone the pool member VSs too as that could be a use case, but that would only make sense if we would also clone the pool members on the remote sites' Controllers too which will get increasingly complex.

One option I will likely implement is a mapping table to allow the user to map from a source VS UUID to a destination VS UUID as a pool member - this wouldn't need any direct interaction with the remote sites' Controllers, but the user would have to supply the source/destination UUIDs...not sure how important this use case is.

python/avi/sdk/samples/clone_vs.py
thebertster: Yes, I will add this and also for the similar message for the clone_vs function.

thebertster: Done.

python/avi/sdk/samples/clone_vs.py
thebertster: For now, this use case seems to meet the needs of the customer that requested it, so can we please push this change and then look at a further enhancement in future?

grastogi23 (approved):

PR #1245, opened 2020-09-14
AV-94889: Skip controller status check on failure option

mayank-avinetworks:


go/session/avisession.go
```diff
@@ -476,8 +476,14 @@ func SetControllerStatusCheckLimits(numRetries, retryInterval int) func(*AviSess
...
-func SetNoControllerStatusCheck(avisess *AviSession) error {
-	avisess.ctrlStatusCheckDisabled = true
+func DisableControllerStatusCheckOnFailure(skipControllerStatusCheck bool) func(*AviSession) error {
+	return func(sess *AviSession) error {
+		return sess.disableControllerStatusCheckOnFailure(skipControllerStatusCheck)
+	}
+}
+
+func (avisess *AviSession) disableControllerStatusCheckOnFailure(skipControllerStatusCheck bool) error {
+	avisess.ctrlStatusCheckDisabled = skipControllerStatusCheck
```
manojkumarjain: Modify Prop name ctrlStatusCheck bool =false

Set in  NewAviSession 
Set  default Value 
ctrlStatusCheck = false

// SetInsecure - Use this for NewAviSession option argument for allowing insecure connection to AviController
func SetCtrlStatusCheck(controllerStatusCheck bool error {
	ctrlStatusCheck = controllerStatusCheck
	return
}

In Test Call the Set Method to Override
aviSession, err = NewAviSession(AVI_CONTROLLER, AVI_USERNAME,
			SetTenant(AVI_TENANT), SetAuthToken(AVI_AUTH_TOKEN), SetInsecure, SetLazyAuthentication(true),    SetCtrlStatusCheck(true), ..........

I think we  just follow simple coding conventions

manojkumarjain (changes requested):
Please Modify

go/session/avisession.go
mayank-avinetworks: Hi Manoj, there are many variables related to controllerStatusCheck.. we need to specify the exact meaning of variable when it will be used. like it is currently depicting that it will disable the controller status check only when there is a failure.

manojkumarjain: Change Variable name to  match the function.  I didn't find any other place where this  flag is used

manojkumarjain (approved):

manojkumarjain (commented):
DisableControllerStatusCheckOnFailure(controllerStatusCheck)

Can we change it to
SetDisableControllerStatusCheckOnFailure(controllerStatusCheck)

sudswasavi:
@manojkumarjain / @mayank-avinetworks I think we will implement this hash in the next release of AKO. It looks like there are still some pending changes being requested and our sys test has run all tests which passed for the hang issue with the earlier hash.

cc @pmanur / @apalsule / @hemantavi / @monotosh-avi

PR #1256, opened 2020-10-16
AV-97907 Modified test case with self signed certificate

yograjshisode:


go/examples/test/custom_transport_test.go
```diff
@@ -15,17 +17,16 @@ func TestCustomTransport(t *testing.T) {
...
+	caCert, _ = ioutil.ReadFile("example.crt")
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caCert)
 
 	transport = &http.Transport{
 		TLSClientConfig: &tls.Config{
+			//Using self signed certificate
 			InsecureSkipVerify: true,
-			// RootCAs:            caCertPool,
+			RootCAs:            caCertPool,
```
mayank-avinetworks: Hi @yograjshisode , Please add a detailed comments, on why self signed certificates are rejected by HTTP client and for that why we skip secure verification. Also, please add a comment saying that if controller has server cert signed from a root CA then this flag should be set to false

mayank-avinetworks (approved):

go/examples/test/custom_transport_test.go
manojkumarjain: No, We should not  InsecureSkipVerify: true
How can we do Hostname verification should happen
We can generate certificate at runtime before test  with correct hostname.

manojkumarjain (approved):

go/examples/test/custom_transport_test.go
```diff
@@ -2,46 +2,165 @@ package test
...
+		DaysUntilExpire: &daysUntilExpire,
+	}
+	sslWithSan := models.SSLKeyAndCertificate{
+		Name: &name,
+		Type: &sslType,
+		Certificate: &certificate,
+	}
+
+	//Create controller session using(using default transport setting)
+	aviClient, err := clients.NewAviClient(os.Getenv("AVI_CONTROLLER"), os.Getenv("AVI_USERNAME"),
```
mayank-avinetworks, line 60: This is still an insecure call. If the intention is just to show that our SDK works well with both self signed ( provided self signed certs are properly generated) and cert auth. signed certs. then I have no issues. Otherwise this does not serve the purpose of end to end secure communication as first few calls are over insecure channel.

yograjshisode: As discussed its just the example to show Go SDK works with custom transport and properly generated certificates.

PR #1259, opened 2020-10-27
Exception error message issue resolved

tusharlimbhore:


chaitanyaavi (approved):

java/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.vmware.avi.sdk</groupId>
 	<artifactId>avisdk</artifactId>
-	<version>20.1.3</version>
```
chaitanyaavi, line 7: version should be 20.1.3

java/pom.xml
```diff
@@ -4,7 +4,7 @@
...
+	<version>20.1.2</version>
```
chaitanyaavi: version should be 20.1.3

PR #1265, opened 2020-10-30
AV-99104 Adding some sleep before creating session with TLS because c…

yograjshisode:
…ontroller takes some time to reflect the systemconfiguration portal cert changes

### DIFF
--- a/go/examples/test/custom_transport_test.go
+++ b/go/examples/test/custom_transport_test.go
@@ -133,6 +133,7 @@ func TestCustomTransport(t *testing.T) {
 		}
 
 	//Try to create the new session with TLS(without InsecureSkipVerify=true)
+	time.Sleep(30 * time.Second)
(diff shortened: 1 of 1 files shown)


go/examples/test/custom_transport_test.go
```diff
@@ -133,6 +133,7 @@ func TestCustomTransport(t *testing.T) {
 		}
 
 	//Try to create the new session with TLS(without InsecureSkipVerify=true)
+	time.Sleep(30 * time.Second)
```
mayank-avinetworks, line 136: Hi @yograjshisode , isnt 30 seconds a lot of time for tests to sleep, usually this is not a very good practice to pause in between tests, but I see there is no way out, but request to look if it can be reduced from 30 secs.

mayank-avinetworks (approved):
Please see if pause time can be reduced.

yograjshisode:
It's safe to have it for 30 sec

PR #1270, opened 2020-11-09
#AV-96433 Added controller password update example to SDK

shounak-opcito:


python/avi/sdk/samples/change_password_example.py
```diff
@@ -0,0 +1,67 @@
...
+    api = None
+
+    try:
+        api = ApiSession(controller_ip=args.controller_ip, username=args.user,
+                         password=args.current_password, tenant="admin")
+    except APIError as error:
+        print("Failed to update the password. Error message: " + str(error))
+        exit(1)
+    change_pwd_example = ChangePasswordExample(api_session=api)
+    change_pwd_example.change_password(args.new_password)
```
gangaadhars, line 83: After changing password and clearing the cached session, user needs to create new api_session with new password. Is that correct.
Can you run some simple get api that needs auth after changing password.

shounak-opcito: Added the get call to get pool before and after updating of the password.

gangaadhars (approved):
Thanks for adding this example.
LGTM.

chaitanyaavi (approved):

PR #1279, opened 2020-11-24
AV-97465: Handle 419 response properly for go sdk.

mayank-avinetworks:
Go SDK delegates control to the controller status poller and also retries to init the session with existing creds in session object after password change which is not a proper way of handling 419 response. 

Once the password is changed he older credentials saved in session object are obsolete and there is no benefit of re initialising the session from SDK with older creds.

```mayank-dev:/home/aviuser/project/go_sdk/src/github.com/avinetworks/sdk/go/session$go test -v
=== RUN   TestAviSession
--- PASS: TestAviSession (2.97s)
=== RUN   TestAviSessionControllerStatusCheckLimits
--- PASS: TestAviSessionControllerStatusCheckLimits (0.00s)
=== RUN   TestAviSessionControllerStatusCheckDisabled
--- PASS: TestAviSessionControllerStatusCheckDisabled (0.00s)
=== RUN   TestAviPool
--- PASS: TestAviPool (1.34s)
=== RUN   TestTenantSwitch
--- PASS: TestTenantSwitch (4.11s)
=== RUN   TestApiLogout
--- PASS: TestApiLogout (4.84s)
=== RUN   TestApiReLogin
--- PASS: TestApiReLogin (5.12s)
=== RUN   TestAviDefaultFields
--- PASS: TestAviDefaultFields (1.36s)
=== RUN   TestTokenAuthRobustness
--- SKIP: TestTokenAuthRobustness (0.00s)
    avisession_test.go:454: SKIPPING as test not running in controller.
=== RUN   TestTokenAuthRobustnessV2
--- PASS: TestTokenAuthRobustnessV2 (0.02s)
=== RUN   TestAviReads
--- PASS: TestAviReads (1.07s)
=== RUN   TestApiLazyAuthentication
--- PASS: TestApiLazyAuthentication (0.21s)
PASS
ok  	github.com/avinetworks/sdk/go/session	21.046s```

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -139,7 +139,7 @@ func (avisess *AviSession) restMultipartFileObjectUploadRequest(verb string, fil
 			return err
 		}
 		retryReq = true
-	} else if resp.StatusCode == 419 || (resp.StatusCode >= 500 && resp.StatusCode < 599) {
+	} else if resp.StatusCode >= 500 && resp.StatusCode < 599 {
 		resp.Body.Close()
 		retryReq = true
 		glog.Infof("Retrying %d due to Status Code %d", retryNum, resp.StatusCode)
@@ -685,7 +685,7 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 				return nil, err
 			}
 			retryReq = true
-		} else if resp.StatusCode == 419 || (resp.StatusCode >= 500 && resp.StatusCode < 599) {
+		} else if resp.StatusCode >= 500 && resp.StatusCode < 599 {
 			resp.Body.Close()
 			retryReq = true
 			glog.Infof("Retrying url: %s; retry: %d due to Status Code %d", url, retry, resp.StatusCode)
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -849,7 +849,7 @@ func (avisess *AviSession) restMultipartUploadRequest(verb string, uri string, f
 			return err
 		}
 		retryReq = true
-	} else if resp.StatusCode == 419 || (resp.StatusCode >= 500 && resp.StatusCode < 599) {
+	} else if resp.StatusCode >= 500 && resp.StatusCode < 599 {
 		resp.Body.Close()
 		retryReq = true
 		glog.Infof("Retrying %d due to Status Code %d", retry, resp.StatusCode)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: go/session/avisession.go [hunk 1]
```diff
@@ -685,7 +685,7 @@ func (avisess *AviSession) restRequest(verb string, uri string, payload interfac
 				return nil, err
 			}
 			retryReq = true
-		} else if resp.StatusCode == 419 || (resp.StatusCode >= 500 && resp.StatusCode < 599) {
+		} else if resp.StatusCode >= 500 && resp.StatusCode < 599 {
```
monotosh-avi, line 688: Is it possible to return the response in case of 401 also ?
With the current change - the caller would get same the error - "Rest request error, returning to caller" for 401 and 5xx, if avisess.disableControllerStatusCheck is true.

mayank-avinetworks: you had concerns about the behaviour after password change of controller user, for which status code returned will be 419. However, we can change this for 401 as well, but I'm not sure about the consequences of breaking existing yet.

monotosh-avi: Yes Mayank. It would be great if we can handle 401 too. The Jira  AV-97465 was for handling 401. Then we found some issues with 419, and I had raised that concern also.

mayank-avinetworks: sure, I'll try to check the corner cases , if any, and will handle this for 401 by tomorrow.

mayank-avinetworks: Hi @monotosh-avi : I tried handling the 401 response as well, but we have failures with existing functionality as this entails overhaul of go sdk. we'll handle 401 resp like 419 in coming time.

monotosh-avi: In that case, we won't be able to use the updated sdk, instead we'll continue to use the sdk from locally modified vendor. We'll be able to use the latest sdk code when we can get response for 401. 
cc @apalsule

sudswasavi: When we say overhaul of SDK, are we referring to existing test suites breaking with the change? @monotosh-avi can you pls share the SDK fork details here where we have modified this?

monotosh-avi: I have made the change in vendor of AKO, not in an sdk branch. For reference we can check the avisession.go changes in this PR:
https://github.com/vmware/load-balancer-and-ingress-services-for-kubernetes/pull/147/files#diff-d6e1226035e649ad6ccc9e7826fc9d77

PR #1295, opened 2020-12-06
Add app-log fetcher that works for more than 10k logs

michaelgranzow-avi:
This tool overcomes the 10k limit we currently have in our application log API.

* As mentioned in the WAF sync the other day, percent_remaining is unreliable
* Contrary to what I said, 'next' handling is correct: if the 10k limit is reached, there is no 'next' link
* The API claims to support ISO 8601 format, but when I use it, the request is rejected (see comment in line 31)

### DIFF
--- a/python/avi/sdk/utils/app_logs/fetch_logs.py
+++ b/python/avi/sdk/utils/app_logs/fetch_logs.py
@@ -17,8 +17,8 @@ Example of fetching vs application logs
 '''
 
 
-def fetch_logs(api_session, tenant, vs_name, start_date, end_date):
-    outfile = open("fetch_logs.json", "w")
+def fetch_logs(api_session, tenant, vs_name, start_date, end_date, outfile):
+    outfile = open(outfile, "w")
     outfile.write("[\n")
     path = "/analytics/logs/"
     num_fetched = 0
@@ -88,7 +88,7 @@ def fetch_logs(api_session, tenant, vs_name, start_date, end_date):
 
     print("Done")
     outfile.write("\n]\n")
-
+    outfile.close()
 
 def compute_date(value):
     my_tz = datetime.datetime.now(datetime.timezone(datetime.timedelta(0))).astimezone().tzinfo
@@ -124,6 +124,8 @@ if __name__ == '__main__':
                         help='user name', default='admin')
     parser.add_argument('-p', '--password',
                         help='password', default='admin')
+    parser.add_argument('-o', '--outfile',
+                        help='File to store resulting JSON array in', default='fetch_logs.json')
 
     args = parser.parse_args()
 
@@ -136,4 +138,4 @@ if __name__ == '__main__':
 
     api_session = ApiSession(args.controller, args.username, args.password,
                              args.tenant)
-    fetch_logs(api_session, args.tenant, args.vs_name, start_date, end_date)
+    fetch_logs(api_session, args.tenant, args.vs_name, start_date, end_date, args.outfile)
--- a/python/avi/sdk/utils/app_logs/fetch_logs.py
+++ b/python/avi/sdk/utils/app_logs/fetch_logs.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+
+# from __future__ import print_function
+import argparse
+import json
+import datetime
+
+from avi.sdk.avi_api import ApiSession
+from requests.packages import urllib3
+
+urllib3.disable_warnings()
+
+
+HELP_STR = '''
+Example of fetching vs application logs
+    fetch_logs.py -c 10.10.25.42 -p xxxx
+    fetch_logs.py -c 10.10.25.42 -p xxxx -s 3600 -e 60
+'''
+
+
+def fetch_logs(api_session, tenant, vs_name, start_date, end_date):
+    outfile = open("fetch_logs.json", "w")
+    outfile.write("[\n")
+    path = "/analytics/logs/"
+    num_fetched = 0
+    num_to_fetch = -1
+    first_line = True
+    while start_date < end_date:
+        # make request for logs [start_date, up_to]:
+        query_options = "virtualservice={}&start={}&end={}".format(
+            vs_name,
+            # The API claims to expect ISO 8601 format, but rejects the request if we use it:
+            # start_date.isoformat(), end_date.isoformat()
+            # instead it seems to be unable to handle time zones. The following works:
+            f"{start_date:%Y-%m-%dT%H:%M:%S.%fZ}", f"{end_date:%Y-%m-%dT%H:%M:%S.%fZ}"
+        )
+        print("Query String: '{}'".format(query_options))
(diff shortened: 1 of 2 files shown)


python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            break
+
+    print("Done")
+    outfile.write("\n]\n")
+    outfile.close()
+
+
+def compute_date(value):
+    my_tz = datetime.datetime.now(datetime.timezone(datetime.timedelta(0))).astimezone().tzinfo
```
chandrak6464, line 108: should we take timezone as optional arg as the script can be run on other hosts as well. Otherwise 'UTC'

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,142 @@
...
+
+        print(" First time stamp:{},\n last time stamp: {}".format(
+            j['results'][0]['report_timestamp'],
+            j['results'][-1]['report_timestamp']))
+
+        num_fetched += len(j['results'])
+
+        # if there's a 'next' hint in the response, use it
+        #
+        # note that due to a current limitation, for page_size = 10k, there
```
chandrak6464, line 71: if the size is less than 10k, (say 5k), will the next be populated correctly. Then no need of +1 us time processing. Also this will reduce load on controller .

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,142 @@
...
+            result = api_session.get(path, tenant=tenant, params=qs)
+            j = result.json()
+            for applog in j['results']:
+                outfile.write(",\n")
+                outfile.write(json.dumps(applog, indent=4))
+            num_fetched += len(j['results'])
+            print("Newly fetched: {}, total: {}".format(len(j['results']), num_fetched))
+        print("No 'next' link, to fetch: {}, fetched: {}".format(num_to_fetch, num_fetched))
+
+        if num_fetched < num_to_fetch:
```
chandrak6464, line 87: also do we need the num_to_fetch and num_fetched conditions at all given then in the edge cases below we might get fetched > to_fetch. Just wondering to do based on last log time stamp check all the time. No need of next logic as well. simple..

chandrak6464 (commented):
Minor comments. Otherwise, looks fine.

python/avi/sdk/utils/app_logs/fetch_logs.py
michaelgranzow-avi: Yes, for count < 10k, 'next' works as expected. The code is correct in this regard, see 'while next in j' below: the time stamp is not changed, but taken as is from the next link (which only increments 'page').

python/avi/sdk/utils/app_logs/fetch_logs.py
michaelgranzow-avi: We need both: 'next' (for page_size < 10k or if the current limitation is fixed) and num_fetched < to_fetch (as long as the 10k limitation exists).

python/avi/sdk/utils/app_logs/fetch_logs.py
michaelgranzow-avi: My assumption here is that the user will provide time values in the time zone of the machine where the script is being run. The controller can be in a different time zone, so we have to provide the time zone to the API. However: that doesn't work, see comment above (I tested with CET).
@grastogi23 can you shed some light on this, please?

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,142 @@
...
+                             'relative: 3600.0 (from 1h before now)', default='3600')
+    parser.add_argument('-e', '--end',
+                        help='End date for fecthing logs. Absolute dates: 2020-12-06, 2020-12-06T09:00:01.137Z or '
+                             'relative: 60.0 (up to 1 minute before now)', default='0')
+    parser.add_argument('-c', '--controller',
+                        help='controller ip', default='127.0.0.1')
+    parser.add_argument('-u', '--username',
+                        help='user name', default='admin')
+    parser.add_argument('-p', '--password',
+                        help='password', default='admin')
```
mirkodziadzka-avi: I would not set a default password. It does not help the user in practice and it makes bad impression for security.

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,142 @@
...
+        j = result.json()
+        if num_to_fetch == -1:
+            # first fetch
+            num_to_fetch = j['count']
+        print("To fetch: {}, in this batch: {}".format(num_to_fetch, len(j['results'])))
+        if num_to_fetch == 0:
+            break
+        if len(j['results']) == 0:
+            print("Expected results, but none found!")
+            exit(1)
```
mirkodziadzka-avi: this will leave the output file in an inconsistent state

mirkodziadzka-avi (commented):
For future development, I would propose to use the logger API instead of print - mainly to distinguish between debug output and error output. And add a --quiet option to the command lines.

python/avi/sdk/utils/app_logs/fetch_logs.py
michaelgranzow-avi: Yes, and anyway it's rather nasty to just 'exit' from a function.

mirkodziadzka-avi: Python style: Define a special exception, raise it here and handle it outside the function (this includes the print)

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -0,0 +1,155 @@
...
+        # if there's a 'next' hint in the response, use it
+        #
+        # note that due to a current limitation, for page_size = 10k, there
+        # is never a 'next' link. this can change, so we handle it here anyway
+        while 'next' in j:
+            # fetch page after page, update num_fetched, save result
+            next = j['next']
+            qs = next.split('?')[1]
+            print("New query string from 'next': {}".format(qs))
+            result = api_session.get(path, tenant=tenant, params=qs)
```
grastogi23, line 78: I was wondering if it would be possible to use response.iter_content() and response.iter_lines() instead. If it makes sense this could be an enhancement for the avi SDK to expose the stream interface.

michaelgranzow-avi: @grastogi23 I'm probably missing some background here, but currently, result.iter_content() simply iterates over the response string byte by byte. Are you suggesting an extension to the API?

chandrak6464 (approved):
Looks good to me. Tested this on our SaaS Ops controller.

PR #1317, opened 2021-01-13
Logs fetcher: combine parser and fetcher, add more options,

michaelgranzow-avi:
e.g. summary-only mode


python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -1,14 +1,17 @@
...
 import json
 import datetime
+from time import sleep
 
 from avi.sdk.avi_api import ApiSession
 from requests.packages import urllib3
 
 urllib3.disable_warnings()
 
+from parse_logs import process_applog_entry, DictOfInts, show_results
```
mirkodziadzka-avi: As our goal is to distribute a script to customers, we should probably inline this for this use case.

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -24,24 +27,38 @@ def __init__(self, num_to_fetch, num_fetched, page_size):
...
+    waf_rules = collections.defaultdict(int)
+    # { "ARGS:foo" => { "count": 137, "values" => { "val1:" 3, "val2": 66 } } }
+    match_elements = {}
+
+    waf_hits = 0
+    waf_elements = 0
+
+    if outfile:
+        outfile = open(outfile, "w")
+        outfile.write("[\n")
```
mirkodziadzka-avi: nit: looks hacky. What is the type of outfile?

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -24,24 +27,38 @@ def __init__(self, num_to_fetch, num_fetched, page_size):
...
+        fixed_options += "&cols={}".format(fields)
     while start_date < end_date:
         # make request for logs [start_date, up_to]:
-        query_options = "virtualservice={}&start={}&end={}".format(
-            vs_name,
+        time_options = "&start={}&end={}".format(
             # The API claims to expect ISO 8601 format, but rejects the request if we use it:
             # start_date.isoformat(), end_date.isoformat()
             # instead it seems to be unable to handle time zones. The following works:
             f"{start_date:%Y-%m-%dT%H:%M:%S.%fZ}", f"{end_date:%Y-%m-%dT%H:%M:%S.%fZ}"
```
mirkodziadzka-avi, line 57: This now requires at least python 3.8 I think. As we give this to a customer, we should check this at the beginning of the script and give a proper error message.

Alternative: ```start_data.format("%Y-%m-%dT%H:%M:%S.%fZ")```

nit: Why it is a start_*date* when it contains date and time?

python/avi/sdk/utils/app_logs/parse_logs.py
```diff
@@ -0,0 +1,227 @@
...
+        for line, row in enumerate(logs, 1):
+            if verbose:
+                print("Processing line {}".format(line))
+            for f in fields_with_structure:
+                if row.get(f, None):
+                    try:
+                        info = ast.literal_eval(row[f])
+                    except Exception as e:
+                        print("Failed to parse column {} in line {} as "
+                              "python AST:\n{}\n Error: ".format(f, line, row[f], e))
```
mirkodziadzka-avi: There is a `{}` missing after `Errror: `

python/avi/sdk/utils/app_logs/fetch_logs.py
mirkodziadzka-avi: As we want to give this to a customer: can we inline this to create a single script?

python/avi/sdk/utils/app_logs/fetch_logs.py
mirkodziadzka-avi: nit: type confusion: What is the type from `outfile`?

python/avi/sdk/utils/app_logs/fetch_logs.py
mirkodziadzka-avi: ~~this requires Python 3.8 or later (I think). As we want to give this to a customer, we should probably document this somewhere. Or we write ```start_data.format("%Y-%m-%dT%H:%M:%S.%fZ}")``` instead.~~

nit: why is it start_DATE if it contains a time?

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -51,14 +68,20 @@ def fetch_logs(api_session, tenant, vs_name, start_date, end_date, outfile):
...
-            first_line = False
-            outfile.write(json.dumps(applog, indent=4))
+            hits, elements = process_applog_entry(applog, uas, browsers, waf_rules, match_elements)
+            waf_hits += hits
+            waf_elements += elements
+
+            if outfile:
+                if not first_line:
+                    outfile.write(",\n")
+                    first_line = False
```
mirkodziadzka-avi: I think `first_line = False` belongs outside this block

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -125,21 +161,29 @@ def compute_date(value):
...
-                        help='End date for fecthing logs. Absolute dates: 2020-12-06, 2020-12-06T09:00:01.137Z or '
+                        help='End date for fetching logs. Absolute dates: 2020-12-06, 2020-12-06T09:00:01.137Z or '
                              'relative before now: 60.0 (up to 60 seconds = 1 minute before now) or 1.5d',
                         default='0')
-    parser.add_argument('-c', '--controller', help='controller ip', default='127.0.0.1')
-    parser.add_argument('-u', '--username', help='user name', default='admin')
+    parser.add_argument('-n', '--top_n', help='Top N occurrences to show', default=10, type=int)
+    parser.add_argument('-f', '--fields', help='Log fields to fetch, e.g. user_agent,client_ip; default: none for all fields',
+                        default=None)
+    parser.add_argument('-o', '--outfile', help='File to store resulting JSON array in. If not supplied, only summary is saved',
```
mirkodziadzka-avi: s/saved/printed/ ?

mirkodziadzka-avi (commented):
If we want to give this to a customer, we should add a README and a `requirements.txt`

python/avi/sdk/utils/app_logs/fetch_logs.py
michaelgranzow-avi: Currently printed, should be saved. Thanks!

mirkodziadzka-avi: How to "save" if there is no outfile?

michaelgranzow-avi:
Guys, we need a better name for this tool - ideas?

python/avi/sdk/utils/app_logs/fetch_logs.py
```diff
@@ -121,35 +378,71 @@ def compute_date(value):
...
+    parser.add_argument('-n', '--top_n', help='Top N occurrences to show', default=10, type=int)
+    parser.add_argument('-o', '--outfile',
+                        help='File to store resulting JSON array in. If not supplied, only summary is saved',
+                        default=None)
     parser.add_argument('-p', '--password', help='password')
-    parser.add_argument('-a', '--authtoken', help='Authentication token')
-    parser.add_argument('-o', '--outfile', help='File to store resulting JSON array in', default='fetch_logs.json')
+    parser.add_argument('-s', '--start',
+                        help='Start date for fetching logs. Absolute dates: 2020-12-06, 2020-12-06T09:00:01.137Z or '
+                             'relative before now: 3600.0 (3600 seconds = 1h before now) or 1d (in the last day)',
```
chandrak6464: Trivial: May be better to organize the help in order. crredentials (controller-ip/token/username/password/tenent) (start,end date) (summary options..) (input, output files..)

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        description=(HELP_STR))
+    parser.add_argument('-c', '--controller', help='controller ip or host name', default='127.0.0.1')
+    parser.add_argument('-i', '--inputfile',
+                        help='File containing log entries - CSV or JSON. If provided controller is not queried')
+
+    parser.add_argument('-u', '--username', help='User name', default='admin')
+    parser.add_argument('-a', '--authtoken', help='Authentication token')
+    parser.add_argument('-p', '--password', help='Password. '
+                        'Deprecated, please use authentication token instead if possible.')
+    parser.add_argument('-t', '--tenant', help='tenant name', default=None)
```
chandrak6464: is default None or 'admin'

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+
+    parser.add_argument('-v', '--vs_name', help='Virtual Service for which AppLogs are fetched')
+
+    parser.add_argument('-b', '--batchsize', help='Batch size for fetching logs', default=10000, type=int)
+    parser.add_argument('-d', '--delay', help='Delay in seconds between requests, eg 0.1', default=1, type=float)
+    parser.add_argument("--dns", action="store_true", help='Make reverse DNS lookup for IPs')
+
+    parser.add_argument('-l', '--logall', help='Log all JSON data received from controller into files under /tmp',
+                        action="store_true")
+    parser.add_argument('--no_ip_obfuscation', help='Do not obfuscate client IPs in all output - not recommended.',
```
chandrak6464, line 577: can this be obfuscate_ip instead of negation?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        # To avoid fetching the same us twice, we increment by 1 us (and risk
+        # missing logs if there is more than 1 per us)
+        start_date = j['results'][-1]['report_timestamp']
+        start_date = datetime.datetime.fromisoformat(start_date) + datetime.timedelta(microseconds=1)
+        if verbose:
+            print("New start date: {}".format(start_date.isoformat()))  # f"{start_date:%Y-%m-%dT%H:%M:%S.%fZ}"))
+        if delay > 0:
+            sleep(delay)
+
+    if num_fetched >= num_to_fetch:
```
chandrak6464: num_fetched is  global across given start and end time, where as num_to_fetch is with in that window of say 8 min. This comparison might not be correct ?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+                print("New query string from 'next': {}".format(qs))
+            if delay > 0:
+                sleep(delay)
+            result = api_session.get(path, tenant=tenant, params=qs)
+            if log_all:
+                logfile_name = "/tmp/analyze_logs_logall-{}.gz".format(logall_filecount)
+                logall_filecount += 1
+                logfile = gzip.open(logfile_name, "wt")
+                logfile.write(result.text)
+                logfile.close()
```
chandrak6464: do you want to use with gzip.open .. for simplified exception handling..similarly at other places ?

chandrak6464: Also why do we need the /tmp intermediate files.  we anyway have complete logs in outfile ?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+
+        success = analyze_logs(api_session,
+                               start_date, end_date,
+                               batch_size, delay,
+                               stats,
+                               args)
+        exit(0 if success else 1)
+    except Exception as e:
+        print(str(e))
+        exit(1)
```
chandrak6464, line 656: should we remove tmp, outfield and analysis files if there is error just to clear the disk/trace..

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+  1 <==> <script>'''
+
+
+def DictOfInts():
+    """Default constructor for a dict of ints
+    """
+    return collections.defaultdict(int)
+
+
+NONCE = urandom(16)
```
mirkodziadzka-avi: I think this need a comment:
- Why we are doing this here (per invocation NONCE to anonymize data)
- That this call can block and what to do here.

I think this need more documentation than the helper class `DictOfInts` above.

Maybe set NONCE = None here and initialize it in `scramble` below. This would also move the potential blocking operation after the start of main(). Another reason to move it down to scramble is that the 16 in urandom and the 16 in hexdigest below is a little bit related. The more details you want in hexdigest, the more entropy you need in the nonce.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    h = sha256()
+    h.update(str(len(s)).encode())
+    h.update(b":")
+    h.update(s.encode())
+    h.update(NONCE)
+    return h.hexdigest()[:16]
+
+
+def load_csv_file(csv_file: str, jsn_file: str, verbose: bool):
+    """Load AVI Application Logs from CSV file 'csv_file'. The Result is
```
mirkodziadzka-avi: s/AVI/Avi/

mirkodziadzka-avi: The name is misleading. It does not load a CSV file. It also has a side effect of writing a JSON file. Why do we need the write here? Can this be done by the called if needed?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    stored as an array in JSON format in a new file 'jsn_file', which will be
+    overwritten if it already exists.
+    The array is returned.
+    """
+    result = []
+    with open(csv_file) as fp:
+        if verbose:
+            print("CSV")
```
mirkodziadzka-avi: for the whole script: maybe `logger.debug` for debugging output?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        logs = csv.DictReader(fp, delimiter=',', quotechar='"')
+        fields_with_structure = [
+            'waf_log', 'significant_log', 'client_cipher_list', 'paa_log', 'icap_log', 'saml_log', 'jwt_log',
+            'DataScriptErrorTrace', 'connection_error_info']
+
+        for line, row in enumerate(logs, 1):
+            if verbose:
+                print("Processing line {}".format(line))
+            for f in fields_with_structure:
+                if row.get(f, None):
```
mirkodziadzka-avi: `None` is not needed here.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        with open(jsn_file, "w") as jsnfile:
+            jsnfile.write(json.dumps(result, sort_keys=True, indent=4))
+    return result
+
+
+def show_top_matchelements(match_elements: dict, N: int, out_fd=sys.stdout):
+    """Writes information about the top 'N' WAF match elements, i.e., the
+    ones most often flagged by WAF, in 'match_elements' to 'out_fd'.
+    For each match elements, the top 'N' match values are also shown.
+    """
```
mirkodziadzka-avi, line 204: General comment. I would separate the computation from the representation. One function for getting the "top N" and one function for printing them.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    else:  # try appending csv or json
+        candidate = filename + ".csv"
+        if path.isfile(candidate):
+            csv_file = candidate
+            jsn_file = filename + ".json"
+        else:
+            candidate = filename + ".json"
+            if path.isfile(candidate):
+                jsn_file = candidate
+    return csv_file, jsn_file
```
mirkodziadzka-avi: Do we need this? I think guessing filenames from user input is not a good thing.

python/avi/sdk/utils/app_logs/analyze_logs.py
mirkodziadzka-avi: by the way: Is reading the json cache really faster? How much? Does it matter?

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    for applog in result:
+        hits, elements = process_applog_entry(applog, uas, stats, waf_rules, match_elements)
+        waf_hits += hits
+        waf_elements += elements
+
+    return uas, waf_rules, match_elements, waf_hits, waf_elements
+
+
+def show_results(top_n: int, obfuscate_ips: bool, dns: bool,
+                 num_entries: int, uas: collections.defaultdict(DictOfInts),
```
mirkodziadzka-avi: can we get away from a type perspective with `Dict[Any, Dict[Any, int]]` here? Because the defaultdict property is not needed here it should work with any dict.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    show_top_uas(uas, top_n, obfuscate_ips, dns, out_fd)
+
+    for stat_name, stat in stats.items():
+        show_top_N(stat, top_n, stat_name, out_fd)
+
+    show_top_N(waf_rules, top_n, 'WAF rules', out_fd)
+    show_top_matchelements(match_elements, top_n, out_fd)
+
+
+def parse_logs(filename: str, stats: dict, N: int, obfuscate_ips: bool, dns: bool, verbose: bool):
```
mirkodziadzka-avi: maybe `use_dns` instead of `dns`. IMHO, type `bool` does not really make sense with a noun/

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    if fields:
+        fixed_options += "&cols={}".format(fields)
+    minutes_tofetch = 8
+    max_minutes = 64
+    min_minutes = 1
+    orig_start_date = start_date
+    while start_date < end_date:
+        upper_end = start_date + datetime.timedelta(minutes=minutes_tofetch)
+        if upper_end > end_date:
+            upper_end = end_date
```
mirkodziadzka-avi: not tested, but aren't these 3 this the same as
```py
upper_end = min(start_date + datetime.timedelta(minutes=minutes_tofetch), end_data)
```

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+
+    result = get_log_data(filename, verbose)
+    uas, waf_rules, match_elements, waf_hits, waf_elements = process_results(result, stats)
+    show_results(N, obfuscate_ips, dns, len(result), uas, stats, waf_rules, match_elements, waf_hits, waf_elements)
+
+
+def analyze_logs(api_session: ApiSession,
+                 start_date: datetime.datetime, end_date: datetime.datetime, page_size: int,
+                 delay: float, stats,
+                 args: dict):
```
mirkodziadzka-avi, line 381: I think this function is very long and hard to read. It intervenes reading the data, analysing the data, writing debug information, writing the result and more. I'm not sure how to simplify this.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        if verbose:
+            print("Query String: '{}'".format(query_options))
+
+        result = api_session.get(path, tenant=tenant, params=query_options)
+        if log_all:
+            logfile_name = "/tmp/analyze_logs_logall-{}.gz".format(logall_filecount)
+            logall_filecount += 1
+            logfile = gzip.open(logfile_name, "wt")
+            logfile.write(result.text)
+            logfile.close()
```
mirkodziadzka-avi: would be more readable if this would be something like:
```py
if log_all:
    logwriter.write(result.text)
```

and logwriter would handle the filecount increment itself

Or
```py
if log_all:
    logall_filecount = logwriter.write(result.text), logall_filecount)
```

In any way, I think there is value in making this output stuff shorter

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+            # It seems to be unable to handle time zones. The following works:
+            f"{start_date:%Y-%m-%dT%H:%M:%S.%fZ}", f"{upper_end:%Y-%m-%dT%H:%M:%S.%fZ}"
+        )
+        query_options = fixed_options + time_options
+        if verbose:
+            print("Query String: '{}'".format(query_options))
+
+        result = api_session.get(path, tenant=tenant, params=query_options)
+        if log_all:
+            logfile_name = "/tmp/analyze_logs_logall-{}.gz".format(logall_filecount)
```
mirkodziadzka-avi: Why are we writing to '/tmp' here?  Do we need these temporary files? 

Anyway, due to security concerns, I would either use "x" as a flag to open below or use a dedicated tempdir

```py
tempdir = tempfile.TemporaryDirectory(dir="/tmp")
...
tempdir.cleanup()
```

(or use it with a context manager in the first place)

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+
+        if verbose:
+            print(" First time stamp:{},\n last time stamp: {}".format(
+                j['results'][0]['report_timestamp'],
+                j['results'][-1]['report_timestamp']))
+
+        num_fetched += len(j['results'])
+        remaining_time = end_date - upper_end
+        r_minutes = round(remaining_time.days * 24 * 60 + remaining_time.seconds / 60)
+        r_percent = round(100 * remaining_time / (end_date - orig_start_date))
```
mirkodziadzka-avi: Do we really want to round here? Or do we want to round down? Just asking, I'm not sure.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+    if num_fetched >= num_to_fetch:
+        print("Done")
+    else:
+        print("Download incomplete: Expected {}, got {} log lines".format(num_to_fetch, num_fetched))
+    if outfile_name:
+        outfile.write("\n]\n")
+        outfile.close()
+
+    if num_fetched > 0:
+        out_fd = gzip.open("analyze_logs.txt.gz", "wt")
```
mirkodziadzka-avi: should this be an argument to the script? If this is a magic filename, I would use "xt" to not override an existing file.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: As discussed the other day, the 'count' is unreliable. Originally num_to_fetch was what we got as 'count' in the first request (with the full window). Essentially, we don't really know how many entries to expect. Will try to find a better way.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: That's what I had at first. But we want to make IP obfuscation the default, so nobody accidentally leaks PII. That's also why there is no short option for this. You should only turn it off if you really know what you're doing.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: From this script's PoV, it's None, but inside ApiSession, None means 'admin'. Will fix.

python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -0,0 +1,632 @@
...
+        if value.endswith('d'):
+            value = float(value[:-1]) * 24 * 3600  # days -> seconds
+        delta = float(value)
+        seconds = int(delta)
+        micros = int((delta-seconds)*100000)
+        result -= datetime.timedelta(seconds=seconds, microseconds=micros)
+    except Exception as e:
+        result = datetime.datetime.fromisoformat(value)
+
+    return result
```
mirkodziadzka-avi, line 535: Are we returning a data object with timezone information or without timezone information? We should not match this. Why do we need the timestamp information in the first place? Is our API using it?

mirkodziadzka-avi (commented):
A lot of nitpick comments, only style questions and remarks

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: This is for debugging to be able to see the full API response, including 'count', 'percent_remaining' etc. Normal users won't need it.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: It's a factor 5 difference if there are 10k lines in the csv. Obviously, for a handful of lines it's negligible.

mirkodziadzka-avi: > It's a factor 5 difference if there are 10k lines in the csv. Obviously, for a handful of lines it's negligible.

What are the absolute numbers?

michaelgranzow-avi: 5s vs 1s.

mirkodziadzka-avi: > 5s vs 1s.

Personal opinion: For a script which is only called manually and very seldom on a system, is it not worth to have this extra json cache to save some seconds per invocation. YMMV

michaelgranzow-avi: I think you are right. It adds too much complexity for the benefit.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: If the user supplies the time as a string in isoformat, the time zone info is optional I believe. So it's up to the user whether they want to include it or not.

If they supply relative times (e.g. 1 hour before now), we must assume that they mean relative to 'now' in their time zone. The controller might live in a different time zone, so we include the TZ of the machine where the script is running.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: We want to display the remaining fraction to the full percent, not more precise. If we round down, it would be more imprecise than necessary, so why do that.

One could argue that we should round *up* because for 0.4% we would show "0% remaining" which sounds stupid.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: According to the docs, such a tempdir will be auto-deleted, but that's not what we want. Will use 
```.py
tempfile.NamedTemporaryFile
```
with deleted=False

chandrak6464: also one more thing. The script will be mostly be executed on controller [ for security/sdk reasons]. We might fill disk which some of our controller processes needs. Also if customer runs the script multiple times, some will get overwritten . I feel better to avoid tmp.

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: No, because in that case we might want to look at those files to understand what went wrong.

python/avi/sdk/utils/app_logs/analyze_logs.py
mirkodziadzka-avi: @michaelgranzow-avi If we want to keep the files, we should let the user specify the location of the files explicitly. Maybe only a single zip file instead of a directory.

michaelgranzow-avi:
@chandrak6464, Regarding overwriting of output files: We never overwrite an existing stats file because that's the main result of an invocation. We do overwrite "side-effect" files like JSON arrays, full API response dumps. These are big and mostly there if someone is interested to dig deep. We don't wan't those big files to accumulate, so I've removed the random tmp-file logic.

michaelgranzow-avi:
@grastogi23 , as discussed, assigning to you for review and approval. I've incorporated all the suggestions from our meeting the other day.

chandrak6464:
@mirkodziadzka-avi is this diff fine ? This is inline with what we discussed in the meeting with broader audience. Any improvements can be done later [ by the relevant teams - say Pulse, if they have specific needs or use-cases for the field].  Would like to merge this so that others can try..

chandrak6464 (approved):

chandrak6464:
< any enhancements can be done with separate tickets based on priority>

PR #1332, opened 2021-02-03
Python dict keys have to be hashable

michaelgranzow-avi:
Don't crash when computing top-n stat for complex field

### DIFF
--- a/python/avi/sdk/utils/app_logs/analyze_logs.py
+++ b/python/avi/sdk/utils/app_logs/analyze_logs.py
@@ -306,9 +306,7 @@ def process_applog_entry(applog: dict, uas: collections.defaultdict(DictOfInts),
 
     for stat_name, stat in stats.items():
         key = applog.get(stat_name, '<NONE>')
-        if type(key) not in ['int', 'float', 'str']:
-            key = str(key)
-        stat[key] += 1
+        stat[str(key)] += 1  # 'str' because the field value could be an array or a dict
 
     waf_log = applog.get('waf_log', "")
     if waf_log != "":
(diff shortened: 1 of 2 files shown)


python/avi/sdk/utils/app_logs/analyze_logs.py
```diff
@@ -305,7 +305,10 @@ def process_applog_entry(applog: dict, uas: collections.defaultdict(DictOfInts),
     uas[ua][applog.get('client_ip', '<UNKNOWN>')] += 1
 
     for stat_name, stat in stats.items():
-        stat[applog.get(stat_name, '<NONE>')] += 1
+        key = applog.get(stat_name, '<NONE>')
+        if type(key) not in ['int', 'float', 'str']:
+            key = str(key)
+        stat[key] += 1
```
mirkodziadzka-avi: @michaelgranzow-avi

1. What is an example for a key which does not work?
2. Can we do the conversion independent of the type?
```py
stats[str(key)] += 1
```

mirkodziadzka-avi (approved):
looks ok

python/avi/sdk/utils/app_logs/analyze_logs.py
michaelgranzow-avi: significant_log is an array, for example (added to the automated test in the makefile).

I had considered doing it unconditionally, but my concern was performance as this will be done for each 'stats' field in every log entry.

Would you say that this is fast enough? If so, I'm happy to change it.

mirkodziadzka-avi: I would not expect performance problems, but you can measure it if needed.

PR #1353, opened 2021-02-24
#AV-102732: Java SDK release script changes

chaitanyaavi:


create_release.sh
```diff
@@ -64,9 +79,23 @@ else
     exit 1
 fi
 
+# Release Avi's Java SDK jar file
+
+if [ -z $AVISDK_PGP_PASSPHRASE ]; then
+  echo "Cannot release java SDK environment is not set"
```
grastogi23: we should say that AVISDK_PGP_PASSPHRASE is not set. So anyone running it knows without looking at the code.

grastogi23 (changes requested):

create_release.sh
```diff
@@ -79,8 +108,12 @@ git push -f origin $REL
 cd ..
 rm -rf avitools
 
+
 /usr/local/bin/hub release $hub_op $assets -F ReleaseNote $REL_TAG
```
grastogi23, line 112: while we are at refactoring can we also check for output of the hub command and test for success. Only if the hub command is successful do rest of the following cleanups

chaitanyaavi: Cleanups are harmless even if the release is not done removing will not error out

grastogi23 (approved):

create_release.sh
grastogi23: we should also exit from here. not just be in else.

create_release.sh
```diff
@@ -64,9 +79,23 @@ else
     exit 1
 fi
 
+# Release Avi's Java SDK jar file
+
+if [ -z $AVISDK_PGP_PASSPHRASE ]; then
+  echo "Cannot release java SDK AVISDK_PGP_PASSPHRASE is not set"
```
grastogi23, line 85: There should be an exit here. There is no reason to continue the release if the java cannot be released.

grastogi23 (changes requested):

manojkumarjain (approved):
Approve

PR #1360, opened 2021-03-19
AV-108714 handle API response for 202 status code

amolopcito:


sabandi (approved):

python/avi/sdk/avi_api.py
```diff
@@ -102,7 +102,7 @@ def json(self):
         and raise Exceptions
         returns the Avi object as a dictionary from rsp.text
         """
-        if self.status_code in (200, 201):
+        if self.status_code in (200, 201, 202):
```
chaitanyaavi: Can we check for 2XX status codes 
```self.status_code>199 and self.status_code<300```

File: python/avi/sdk/avi_api.py [hunk 2]
```diff
@@ -102,7 +102,7 @@ def json(self):
         and raise Exceptions
         returns the Avi object as a dictionary from rsp.text
         """
-        if self.status_code in (200, 201):
+        if self.status_code > 199 and self.status_code < 300:
```
sabandi, line 105: with this change  line no 111 becomes redundant.

gangaadhars (approved):
LGTM. Thx

chaitanyaavi (approved):

PR #1364, opened 2021-03-24
#AV-64696 Added new patch syntax

amolopcito:


sabandi:
@amolopcito  can you add the test report this change and also some examples for usage.

PR #1485, opened 2021-07-13
Auto updated assets for python eng

yograjshisode:


python/avi/sdk/samples/clone_vs.py
```diff
@@ -800,8 +800,6 @@ def _process_sslkeyandcertificate(self, ss_obj, t_obj, ot_obj, oc_obj,
                 ss_obj.pop('key_base64', None)
                 ss_obj.pop('certificate_base64', None)
```
chaitanyaavi: This needs to be present @yograjshisode Can you please check with Original PR?

PR #1558, opened 2021-08-18
Auto updated assets for python 21.1.2

yograjshisode:


python/version.py
```diff
@@ -8,7 +8,7 @@
 import argparse
 
 AVI_VERSION = '21.1.1'
```
chaitanyaavi: Version should be 21.1.2

chaitanyaavi (changes requested):
Version not updated for branch

chaitanyaavi (approved):

PR #1628, opened 2021-09-06
NetScaler GSLB fixes

nblagoev:
Couple of bug fixes related to the NetScaler GSLB conversion
- Respect the tenant when fetching the vip cluster map in the NetScaler GSLB conversion
- Fix the 'enabled' flag value when converting NetScaler GSLB
- Remove the meta obj reference from the NetScaler GSLB as it is no longer available in the superclass


chaitanyaavi:
Multi-tenancy in Netscaler not supported state can be changed from params

PR #1750, opened 2021-10-01
Allow for callers to inject their own HTTP client

aaronjwood:
Would be happy to discuss this more next week in our meeting with Manish.

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -334,6 +334,13 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 	}
 
 	if avisess.client == nil {
+		// create default transport object
+		if avisess.transport == nil {
+			avisess.transport = &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			}
+		}
+
 		// attach transport object to client
 		avisess.client = &http.Client{
 			Transport: avisess.transport,
@@ -510,6 +517,9 @@ func SetTransport(transport *http.Transport) func(*AviSession) error {
 }
 
 func (avisess *AviSession) setTransport(transport *http.Transport) error {
+	if avisess.client != nil {
+		return errors.New("Cannot set custom Transport for external clients")
+	}
 	avisess.transport = transport
 	return nil
 }
@@ -522,6 +532,9 @@ func SetClient(client HttpClient) func(*AviSession) error {
 }
 
 func (avisess *AviSession) setClient(client HttpClient) error {
+	if avisess.transport != nil {
+		return errors.New("Cannot set custom client when transport is already set to http.Transport")
+	}
 	avisess.client = client
 	return nil
 }
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -321,13 +321,6 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 		avisess.api_retry_interval = DEFAULT_API_RETRY_INTERVAL
 	}
 
-	// create default transport object
-	if avisess.transport == nil {
-		avisess.transport = &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-		}
-	}
-
 	// set default timeout
 	if avisess.timeout == 0 {
 		avisess.timeout = DEFAULT_API_TIMEOUT
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -53,6 +53,11 @@ type AviError struct {
 	err error
 }
 
+// HttpClient allows callers to inject their own implementations for the SDK to use.
+type HttpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 // PostMultipartRequest performs a POST API call and uploads multipart data to API fileobject/upload
 func (avisess *AviSession) PostMultipartWafAppSignatureObjectRequest(fileLocPtr *os.File, uri string, tenant string, fileParams map[string]string) error {
 	url := avisess.prefix + "/api/wafapplicationsignatureprovider/" + uri
@@ -250,7 +255,7 @@ type AviSession struct {
 	transport *http.Transport
 
 	// internal: reusable client
-	client *http.Client
+	client HttpClient
 
 	// optional lazy authentication flag. This will trigger login when the first API call is made.
 	// The authentication is not performed when the Session object is created.
@@ -328,10 +333,12 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 		avisess.timeout = DEFAULT_API_TIMEOUT
 	}
 
-	// attach transport object to client
-	avisess.client = &http.Client{
-		Transport: avisess.transport,
-		Timeout:   avisess.timeout,
+	if avisess.client == nil {
+		// attach transport object to client
+		avisess.client = &http.Client{
+			Transport: avisess.transport,
+			Timeout:   avisess.timeout,
+		}
 	}
 
 	if !avisess.lazyAuthentication {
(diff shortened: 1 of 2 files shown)


manojkumarjain:
Isn't it injecting clientfactory will be better
It allow you to manage Session and httpclient with your scheme
In GoSDK we have given that features
Similar feature is develop in JavaSDK

aaronjwood:
> Isn't it injecting clientfactory will be better It allow you to manage Session and httpclient with your scheme In GoSDK we have given that features Similar feature is develop in JavaSDK

Sorry, I didn't quite follow you. I don't see a ClientFactory (or Factory) reference anywhere in the Go codebase.

mayank-avinetworks:
Hi @aaronjwood , We support setting custom transport and timeout while creating the avi session in GO SDK.
https://github.com/vmware/alb-sdk/blob/eng/go/examples/test/custom_transport_test.go#L143
I believe this example should solve your use case.

aaronjwood:
> Hi @aaronjwood , We support setting custom transport and timeout while creating the avi session in GO SDK.
> https://github.com/vmware/alb-sdk/blob/eng/go/examples/test/custom_transport_test.go#L143
> I believe this example should solve your use case.

The transport isn't enough to cover our use case. We can discuss why on the call later today.

go/session/avisession.go
```diff
@@ -250,7 +255,7 @@ type AviSession struct {
 	transport *http.Transport
 
 	// internal: reusable client
-	client *http.Client
+	client HttpClient
```
grastogi23, line 258: It has been a while but there was a reason for using a ptr as some optimizations where being done by the core Go HTTP client module. Is there any reason why HTTPClient needs to be passed by value .

How about leave the client as *http.Client. But you can always set it in the  NewAviSession based on whether it was passed in the arguments list.

go/session/avisession.go
```diff
@@ -328,10 +333,12 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 		avisess.timeout = DEFAULT_API_TIMEOUT
 	}
 
-	// attach transport object to client
-	avisess.client = &http.Client{
-		Transport: avisess.transport,
-		Timeout:   avisess.timeout,
+	if avisess.client == nil {
```
grastogi23, line 329: This code works as is. we don't need to make line 258 client HttpClient from client *HttpClient. It also avoids confusion whether the object's HTTPClient is used or one that was passed down.

grastogi23 (changes requested):

go/session/avisession_test.go
```diff
@@ -685,3 +686,30 @@ func TestApiLazyAuthentication(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestSetClient(t *testing.T) {
```
grastogi23, line 691: Thanks for adding this.

go/session/avisession.go
```diff
@@ -507,6 +514,18 @@ func (avisess *AviSession) setTransport(transport *http.Transport) error {
 	return nil
 }
 
+// SetClient allows callers to inject their own HTTP client.
+func SetClient(client HttpClient) func(*AviSession) error {
```
grastogi23, line 521: similar comment here. let it be (client *HttpClient) instead of pss by value

go/session/avisession.go
aaronjwood: Interfaces are references underneath in Go which is why I kept it this way. I haven't seen people often taking pointers to interfaces unless you have some fancy stuff going on. It'd be a pointer to a pointer in that case.

I don't think I fully understood your comment about keeping it as a concrete struct type. If it's not an interface then passing it in `NewAviSession` will only allow for the standard lib `*http.Client` implementation to be used. We want to be able to swap the client out completely, not be restricted by it.

go/session/avisession.go
aaronjwood: If the nil check is omitted then won't it overwrite the custom implementation being passed in? I think I saw the option parsing code above this area, so if you pass in `SetClient` in `NewAviSession` it'll be taken, but then here it'll just be overwritten if there's no guardrail.

go/session/avisession.go
aaronjwood: Are you suggesting we move to using double pointers under the hood? The goal is to move away from a concrete struct type here so I'm relying on the implicit reference that the interface carries.

go/session/avisession.go
grastogi23: Hi
I have limited Go experience. But as far as I know Go is always pass by value - https://medium.com/@vivek_syngh/pass-by-value-and-pass-by-reference-in-go-8483534b7385 (this was based on 5 minute search on the golang)

Regardless, I do not agree with you that a field x y is same as x *y in a class. Why create the confusion? Can you please explain why having client *HttpClient would not work for you?

go/session/avisession.go
grastogi23: I didn't mean remove the nil check. Your code for checking for avisess.client == nil  will work even when you change to client *HttpClient. The code will need to be modified that if client was not passed in the args then supply it &http.Client. If it was then set it to the one passed in the args.

go/session/avisession.go
grastogi23: Making it pointer does not mean it becomes pointer to pointer. It is value of pointer. It keeps the interface clean as lifecyle of the httpclient is outside of the sdk and it is maintained by the caller.

go/session/avisession.go
aaronjwood: No worries. We can make it a pointer to a pointer if needed, I just didn't see the case for it. You're right that Go is pass by value, but in the same way that C is pass by value. Even though you have pointer types they can be passed around by value. Usually with C, or I guess Go if you had the use case, you work with double pointers when you want to change what it's pointing to (because of the constraints of pass by value)

What would be the benefit to having the field (now an interface) be a pointer to a pointer here?

go/session/avisession.go
```diff
@@ -328,10 +333,19 @@ func NewAviSession(host string, username string, options ...func(*AviSession) er
 		avisess.timeout = DEFAULT_API_TIMEOUT
 	}
 
-	// attach transport object to client
-	avisess.client = &http.Client{
-		Transport: avisess.transport,
-		Timeout:   avisess.timeout,
+	if avisess.client == nil {
+		// create default transport object
+		if avisess.transport == nil {
```
grastogi23, line 331: Please delete the lines 325 - 329 337-342 take care of that scenario.

grastogi23 (changes requested):

grastogi23 (approved):

mayank-avinetworks (approved):

mayank-avinetworks:
run jenkins-go-sdk-ut

mayank-avinetworks:
run jenkins-go-sdk-ut

ArpitAVI:
run jenkins-go-sdk-ut

ArpitAVI:
run jenkins-go-sdk-ut

ArpitAVI:
run jenkins-go-sdk-ut

ArpitAVI:
run jenkins-go-sdk-ut

go/session/avisession_test.go
```diff
@@ -685,3 +686,30 @@ func TestApiLazyAuthentication(t *testing.T) {
...
+	var avisess *AviSession
+	var err error
+	client := &stdhttp.Client{Timeout: 10 * time.Second}
```
mayank-avinetworks: @aaronjwood , Please set the transport and timeout for the external client by yourself here.

mayank-avinetworks (approved):

yograjshisode:
run jenkins-go-sdk-ut

mayank-avinetworks:
UT passed for TCs

PR #1804, opened 2021-10-26
#AV-128665: Java SDK login failure throws IOException instead of Http…

chaitanyaavi:
…ClientErrorException in http error


manojkumarjain (approved):
Approve.
Maybe in future we should create our own exception class

sagarpsalvi (commented):
Looks good to me

PR #2071, opened 2022-04-20
Add params argument to _get_uuid_by_name

mslocrian:
The `delete_by_name()` and `put_by_name()` methods of `ApiSession()` accept the `params` argument; however, they do not actually use it in the `_get_uuid_by_name()` call. This can be problematic if you have two objects with the same name that exist in different clouds.

This PR adds params to those calls and associated methods.


mslocrian:
@gangaadhars / @chaitanyaavi / @manojkumarjain , where would be the best place to add the test results that the sdk is failing on?

gangaadhars (approved):

PR #2205, opened 2022-07-01
Added changes to avoid cookies collection

yograjshisode:


go/session/avisession.go
```diff
@@ -630,7 +624,7 @@ func (avisess *AviSession) newAviRequest(verb string, url string, payload io.Rea
 		req.Header.Set("X-Avi-Tenant", tenant)
 	}
 
-	if avisess.sessionid != "" {
+	if !strings.HasSuffix(url, "login") && avisess.sessionid != "" {
```
mayank-avinetworks, line 627: @yograjshisode , what are the consequences of sending this with login URI, I believe the server should ignore all these values if these are sent with Login API, isn't this the case?

yograjshisode: @mayank-avinetworks  It returns error 401 with error "CSRF token missing or incorrect." if we set it with wrong value.

PR #2686, opened 2023-06-13
AV-180818 Fixed F5 vip issue

Rohan-sss1:


parimanur (approved):

parimanur (changes requested):
Are the arguments always in the same order? If they change then the code will break.

Rohan-sss1:
@parimanur this is default format of referencing and tenant and cloud are mandatory while converting so this format will never change

parimanur (approved):

parimanur:
> @parimanur this is default format of referencing and tenant and cloud are mandatory while converting so this format will never change

Can you address the changes in avi-dev?

PR #2717, opened 2023-06-30
AV-183030: Nsxt tool fix 22.1.4

agarwalme:
Pr consist of following fixes:
1. fixed config discovery phase for out of scope error
2. Nsxt cipher conversion


python/avi/migrationtools/nsxt_converter/nsxt_util.py
```diff
@@ -582,20 +583,24 @@ def get_lb_services_details(self):
...
+                     }
+                    LOG.debug("Lb skipped : %s reason %s" %(lb["id"],lb_skip_reason))
+                    continue
+
             if not is_cloud_configured:
 
                 warning_mesg="cloud is not configured for load balancer %s with id %s " % (lb["display_name"],lb["id"])
                 LOG.debug(warning_mesg)
-                lb_details=[]
+               # lb_details=[]
```
parimanur: can you delete this.

python/avi/migrationtools/nsxt_converter/nsxt_util.py
```diff
@@ -125,10 +125,11 @@ def get_object_segments(vs_id, obj_ip):
     if not vs:
         LOG.debug("virtual service not found with id %s " % vs_id)
         return None
-    cloud = vs.get("cloud")
+    cloud = vs.get("Cloud")
+
     if cloud == "Cloud Not Found":
         LOG.debug("cloud is not configured for vs %s " % vs_id)
-        return None
+       # return None
```
parimanur, line 132: why are we proceeding even if we are not able to get the cloud?

agarwalme: We are removing this statements , so that it will check for segments range and provide valid reasons for the pool attached to that vs.

parimanur (approved):

PR #2863, opened 2024-01-19
AV-197483: Added knob to skip datapath validations

shardullatkar:
- Added script parameter to override the behaviour for validating data path. Default is true and if parameter is provided then false (skip validation).
- Tested this with setup with pool members not falling in network ip ranges. With this flag set the pools will be anyway migrated.


agarwalme:
Update readme file as well.

PR #2882, opened 2024-02-06
AV-197650: Update public repo readme to point to new migration tool location

agarwalme:


mkhachane:
Added changes in README file.

PR #3085, opened 2025-06-30
Auto updated assets for java 31.1.2

avijenkins:


mkhachane:
Not needed as we have updated release version

PR #3245, opened 2025-10-28
 Fix: Include session tokens in all the request headers

Rohan-sss1:
Added fix to update the request headers for all the requests

Identified a crucial flaw in the current logic. The strings.HasSuffix(url, "login") check is
indeed too broad. If we have other API endpoints like `/api/user/login-history` or `/api/admin/force-login` that happen to end with "login", the current condition
!strings.HasSuffix(url, "login") will evaluate to false for these, preventing the CSRF token from
being added. This is problematic if those endpoints require a CSRF token for security.

The intent is to only exclude the primary authentication endpoint (which is avisess.prefix + "login") from receiving a CSRF token.

To fix this, I have changed the condition in newAviRequest to specifically check if the url is exactly the login URL, rather than just checking if it ends with "login".

Explanation of the change:

   * `url != avisess.prefix+"login"`: This new part of the condition precisely checks if the current
     request's full URL is not the exact login endpoint.
   * This ensures that:
       * The actual login endpoint (e.g., https://controller.com/login) will correctly not have a
         CSRF token added.
       * Any other API endpoint, even if its path contains or ends with "login" (e.g.,
         https://controller.com/api/user/login-history), will have the CSRF token added (provided
         avisess.csrfToken is not empty), as these endpoints are likely part of the authenticated
         session and require the token.

  This modification makes the CSRF token handling more accurate and prevents unintended exclusion of
  valid API calls.

Testing done by running all the avisession related test cases.
Tested the fix on GSLB setup as well by @arihantg

### DIFF
--- a/go/session/avisession.go
+++ b/go/session/avisession.go
@@ -759,7 +759,7 @@ func (avisess *AviSession) newAviRequest(verb string, url string, payload io.Rea
 	if tenant == "" {
 		tenant = avisess.tenant
 	}
-	if !strings.HasSuffix(url, "login") && avisess.csrfToken != "" {
+	if url != avisess.prefix+"login" && avisess.csrfToken != "" {
 		req.Header["X-CSRFToken"] = []string{avisess.csrfToken}
 		req.AddCookie(&http.Cookie{Name: "csrftoken", Value: avisess.csrfToken})
 	}
@@ -768,7 +768,7 @@ func (avisess *AviSession) newAviRequest(verb string, url string, payload io.Rea
 		req.Header.Set("X-Avi-Tenant", tenant)
 	}
 
-	if !strings.HasSuffix(url, "login") && avisess.sessionid != "" {
+	if url != avisess.prefix+"login" && avisess.sessionid != "" {
 		req.AddCookie(&http.Cookie{Name: "sessionid", Value: avisess.sessionid})
 		req.AddCookie(&http.Cookie{Name: "avi-sessionid", Value: avisess.sessionid})
 	}


arihantg (approved):
LGTM

go/session/avisession.go
```diff
@@ -759,7 +759,7 @@ func (avisess *AviSession) newAviRequest(verb string, url string, payload io.Rea
 	if tenant == "" {
 		tenant = avisess.tenant
 	}
-	if !strings.HasSuffix(url, "login") && avisess.csrfToken != "" {
+	if url != avisess.prefix+"login" && avisess.csrfToken != "" {
```
arihantg, line 762: should this be /api/login

parimanur, line 762: can you add comment on what is the prefix value? Does it have /api in it?
Also can you check login flow as well.

parimanur (approved):
Add the testing for the login workflow.

go/session/avisession.go
Rohan-sss1: this is the prefix https://100.65.8.114/

this is for login api
https://100.65.8.114/login

Rohan-sss1:
> Add the testing for the login workflow.

Test result of all the session related tests
```
\=== RUN   TestAviSession
--- PASS: TestAviSession (3.23s)
=== RUN   TestAviSessionControllerStatusCheckLimits
--- PASS: TestAviSessionControllerStatusCheckLimits (0.00s)
=== RUN   TestAviSessionControllerStatusCheckDisabled
--- PASS: TestAviSessionControllerStatusCheckDisabled (0.00s)
=== RUN   TestAviPool
--- PASS: TestAviPool (1.25s)
=== RUN   TestTenantSwitch
--- PASS: TestTenantSwitch (3.68s)
=== RUN   TestApiLogout
--- PASS: TestApiLogout (3.16s)
=== RUN   TestApiReLogin
--- PASS: TestApiReLogin (3.51s)
=== RUN   TestAviDefaultFields
--- PASS: TestAviDefaultFields (1.33s)
=== RUN   TestTokenAuthRobustness
    avisession_test.go:460: SKIPPING as test not running in controller.
--- SKIP: TestTokenAuthRobustness (0.00s)
=== RUN   TestTokenAuthRobustnessV2
--- PASS: TestTokenAuthRobustnessV2 (0.01s)
=== RUN   TestAviReads
--- PASS: TestAviReads (0.83s)
=== RUN   TestApiLazyAuthentication
--- PASS: TestApiLazyAuthentication (0.22s)
=== RUN   TestSetClient
--- PASS: TestSetClient (0.21s)
=== RUN   TestRawAPI
--- PASS: TestRawAPI (3.76s)
=== RUN   TestAviSessionGoRoutine
--- PASS: TestAviSessionGoRoutine (53.38s)
=== RUN   TestIPV6_V6
--- PASS: TestIPV6_V6 (0.00s)
=== RUN   TestIPV6_V4
--- PASS: TestIPV6_V4 (0.00s)
PASS
ok  	github.com/vmware/alb-sdk/go/session	74.560s
```

